PR #8, opened 2017-10-04
Add a script to check for missing dependencies in asdf-systems/package.xmls

cpollok:


asdf_check.py
```diff
@@ -0,0 +1,206 @@
...
+])
+
+FORMAT_2_DEPEND_TAGS = set([
+    "depend",
+    # "buildtool_depend",
+    "build_depend",
+    "exec_depend",
+    "test_depend",
+    "doc_depend"
+])
```
gaya-: Could you please only use "build_depend" and "depend" tags, as "run_depend" aka "exec_depend" are only run-time dependencies and are only for launch files not the Lisp code. "test_depend" and "doc_depend" are also not for Lisp code, hence irrelevant.

cpollok: Done.

PR #17, opened 2017-10-26
Add retries with different torso angles to low-level move-tcp [DON'T MERGE YET]

cpollok:
Changes the call-ik-solver into a recursive function which recurses until all torso angles in the range from the upper to the lower limit (in steps of 0.01m) have been tried and no call returned a IK solution.
It also prevents the service call from printing a error message by temporarily increasing the debug level.
Otherwise for every failed call a pretty long error message would clutter the REPL.

### DIFF
--- a/cram_pr2/cram_pr2_projection/src/ik.lisp
+++ b/cram_pr2/cram_pr2_projection/src/ik.lisp
@@ -156,10 +156,9 @@
 
 (defparameter *torso-step* 0.01)
 
-(defun call-ik-service (left-or-right cartesian-pose &key seed-state test-angle torso-angle)
+(defun call-ik-service (left-or-right cartesian-pose &key seed-state test-angle torso-angle torso-lower-limit torso-upper-limit)
   (declare (type keyword left-or-right)
            (type cl-transforms-stamped:pose-stamped cartesian-pose))
-  ;; (setf *ee-poses* (cons cartesian-pose *ee-poses*))
   (let ((ik-link (get-ik-solver-link left-or-right))
         (dbg-lvl (roslisp:debug-level NIL))) ;; TODO: have to get a real name?
     (roslisp:set-debug-level NIL 9)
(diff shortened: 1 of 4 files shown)


cram_pr2/cram_pr2_projection/src/ik.lisp
```diff
@@ -154,48 +154,91 @@
...
-                                              response-error-code))))))
+                                  :format-control "IK service failed: ~a"
+                                  :format-arguments (list
+                                                     (roslisp-msg-protocol:code-symbol
+                                                      'moveit_msgs-msg:moveiterrorcodes
+                                                      response-error-code))))))
       (simple-error (e)
         (declare (ignore e))
-        (format t "IK service call freaked out. No IK solution found.~%")))))
+        (cut:with-vars-bound (?upper-limit ?lower-limit)
```
gaya-: Lisp prolog calls are quite expensive in the current Prolog implementation, let's try to avoid unnecessary calls. So 2 solutions for this:

* implement a caching mechanism to not call prolog if limits are already known
* pass the limit as an argument to call_ik

I would prefer 2nd solution.

cram_pr2/cram_pr2_projection/src/low-level.lisp
```diff
@@ -266,8 +266,13 @@
 
 (defparameter *gripper-length* 0.2 "PR2's gripper length in meters, for calculating TCP -> EE")
 
+(defparameter *l-tcp-p* nil)
+(defparameter *r-tcp-p* nil)
+(defparameter *ee-poses* nil)
```
gaya-: So this is not used anymore, was is just for debugging?

cram_pr2/cram_pr2_projection/src/ik.lisp
cpollok: Implemented the second one.

cram_pr2/cram_pr2_projection/src/low-level.lisp
cpollok: Debugging stuff is removed.

gaya-:
Ok, last thing: please export the torso step variable in package.lisp, then I would like to merge this PR.

cpollok:
Exported the torso step variable.

gaya-:
Oh... There was nearly nothing exported from this package at all... Oh well.

gaya-:
Manually merged into cram5 branch: https://github.com/cram2/cram/commit/08ea4221c0ec92ef38a227ec268c48dfb9eb4dd1

gaya-:
Thanks @cpollok for providing the new feature.

PR #31, opened 2017-11-27
logged position parameter

seba90:


cram_core/cram_executive/src/perform.lisp
```diff
@@ -179,6 +179,7 @@ similar to what we have for locations.")
 
 (defun log-action-parameter (designator action-id)
   (cond ((desig-prop-value designator :effort) (ccl::send-effort-action-parameter action-id (write-to-string (desig-prop-value designator :effort))))
+        ((desig-prop-value designator :position) (ccl::send-position-action-parameter action-id (desig-prop-value designator :position)))
```
gaya-: There is no "position" field in the designators...
Don't see how this can work.

PR #37, opened 2018-01-10
[pr2-projection] Fix torso angle resampling in ik solver call.

cpollok:
Fixes the torso angle resampling in the low level ik-solver-call.
Now it resamples in all cases where there's no ik solution.
It also turns call-ik-service back into the pure ROS call it once was.

### DIFF
--- a/cram_pr2/cram_pr2_projection/src/ik.lisp
+++ b/cram_pr2/cram_pr2_projection/src/ik.lisp
@@ -226,7 +226,7 @@
       (unwind-protect
            (progn
              (roslisp:set-debug-level NIL 9)
-             (_call-ik-service-with-torso-resampling left-or-right cartesian-pose :seed-state seed-state
+             (call-ik-service-with-torso-resampling-inner left-or-right cartesian-pose :seed-state seed-state
                                                                                   :test-angle test-angle
                                                                                   :torso-angle torso-angle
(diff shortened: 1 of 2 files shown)


cram_pr2/cram_pr2_projection/src/ik.lisp
```diff
@@ -154,82 +153,85 @@
...
+                                              response-error-code))))))
+      (simple-error (e)
+        (declare (ignore e))
+        nil))))
 
+(defparameter *torso-step* 0.01)
 
+(defun call-ik-service-with-torso-resampling (left-or-right cartesian-pose
+                                              &key seed-state test-angle torso-angle torso-lower-limit torso-upper-limit)
+  (labels ((_call-ik-service-with-torso-resampling (left-or-right cartesian-pose
```
gaya-: Thanks for providing the PR!

Could you please rename ``_call-ik-service-with-torso-resampling`` into ``call-ik-service-with-torso-resampling-inner`` or so? Underscores are very awkward in Lisp, usually inner functions of recursion are named with ``inner`` suffix.

cpollok: Done

gaya-:
Manually merged into ease-demo branch: 1c2fc1c1d361374c6349f0c93198d699fb87a37e

PR #41, opened 2018-02-28
[pr2-proj] Add virtual links in the PR2 URDF to the projection tf tree.

cpollok:
Basically iterate over the difference in links of the URDF and in the bullet world until they are the same.
Achieve this by setting transforms from the virtual link's parents to them and adding them to the bullet world list of links. Repeat until every virtual link is asserted into the projection tf. This works because those virtual links who have other virtual links as parents will be asserted on the next loop iteration.

If there are virtual links who don't exist in the hierarchy of the projection's tf the loop returns to not run into an endless loop. But even then every other virtual link will be asserted.

### DIFF
--- a/cram_pr2/cram_pr2_projection/src/tf.lisp
+++ b/cram_pr2/cram_pr2_projection/src/tf.lisp
@@ -86,16 +86,17 @@
              (urdf-link-names (remove-if (lambda (x) (member x bullet-links))
                                          (loop for name being the hash-keys in urdf-links
                                                collecting name))))
-        (let ((prev-link-count))
+        (let ((prev-link-count 0))
           ;; while there are still links left to be asserted in tf
           (loop while (> (length urdf-link-names) 0)
                 do ;; if no links were asserted last loop, we are stuck and have to return
-                   (if (eq prev-link-count (length urdf-link-names))
+                   (if (= prev-link-count (length urdf-link-names))
                        (progn
-                           (roslisp:ros-error (pr2-proj tf) "Somehow a virtual link with no parents in the bullet world was found in the urdf. This should never happen.")
+                           (roslisp:ros-error (pr2-proj tf) "Somehow a virtual link with no parents in ~
+						   the bullet world was found in the urdf. This should never happen.")
                            (return))
                        (setf prev-link-count (length urdf-link-names)))
-                   (let ((next-round (list)))
+                   (let (next-round)
                      ;; try to assert a transform from a parent in the bullet world to the virtual links
                      (loop for link-name in urdf-link-names
                            do (let* ((link-joint (cl-urdf:from-joint (gethash link-name urdf-links)))
@@ -103,7 +104,8 @@
                                 (when (and (not (eq (cl-urdf:joint-type link-joint) :FIXED))
                                            display-warnings)
                                   (roslisp:ros-warn (pr2-proj tf)
-                                                    "Joint of ~a is ~a. Only links on FIXED joints should be asserted as virtual links."
+                                                    "Joint of ~a is ~a. Only links on FIXED joints should ~
+													be asserted as virtual links."
                                                     (cl-urdf:name link-joint)
                                                     (cl-urdf:joint-type link-joint)))
                                 (if (member parent-name bullet-links)
--- a/cram_pr2/cram_pr2_projection/src/tf.lisp
+++ b/cram_pr2/cram_pr2_projection/src/tf.lisp
@@ -93,7 +93,7 @@
                    (if (= prev-link-count (length urdf-link-names))
                        (progn
                            (roslisp:ros-error (pr2-proj tf) "Somehow a virtual link with no parents in ~
-						   the bullet world was found in the urdf. This should never happen.")
+                           the bullet world was found in the urdf. This should never happen.")
                            (return))
                        (setf prev-link-count (length urdf-link-names)))
                    (let (next-round)
@@ -105,7 +105,7 @@
                                            display-warnings)
                                   (roslisp:ros-warn (pr2-proj tf)
                                                     "Joint of ~a is ~a. Only links on FIXED joints should ~
-													be asserted as virtual links."
+                                                    be asserted as virtual links."
                                                     (cl-urdf:name link-joint)
                                                     (cl-urdf:joint-type link-joint)))
                                 (if (member parent-name bullet-links)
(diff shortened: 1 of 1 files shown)


gaya-:
Great! Thanks. I'll test it next week.

cram_pr2/cram_pr2_projection/src/tf.lisp
```diff
@@ -43,43 +44,81 @@
...
+             (urdf-link-names (remove-if (lambda (x) (member x bullet-links))
+                                         (loop for name being the hash-keys in urdf-links
+                                               collecting name))))
+        (let ((prev-link-count))
+          ;; while there are still links left to be asserted in tf
+          (loop while (> (length urdf-link-names) 0)
+                do ;; if no links were asserted last loop, we are stuck and have to return
+                   (if (eq prev-link-count (length urdf-link-names))
+                       (progn
+                           (roslisp:ros-error (pr2-proj tf) "Somehow a virtual link with no parents in the bullet world was found in the urdf. This should never happen.")
```
gaya-: Please make this line shorter. You can use tilde for this:
```
    "something ~
     and something else"
```

cram_pr2/cram_pr2_projection/src/tf.lisp
```diff
@@ -43,43 +44,81 @@
...
+             (urdf-links (cl-urdf:links (btr:urdf ?robot-instance)))
+             ;; get the names of the links, which are only present in the urdf
+             (urdf-link-names (remove-if (lambda (x) (member x bullet-links))
+                                         (loop for name being the hash-keys in urdf-links
+                                               collecting name))))
+        (let ((prev-link-count))
+          ;; while there are still links left to be asserted in tf
+          (loop while (> (length urdf-link-names) 0)
+                do ;; if no links were asserted last loop, we are stuck and have to return
+                   (if (eq prev-link-count (length urdf-link-names))
```
gaya-: Please use `=` for numbers. For example, `(eq 1.0 1)` is `NIL`, and with `=` it is true.

cram_pr2/cram_pr2_projection/src/tf.lisp
```diff
@@ -43,43 +44,81 @@
...
+                       (progn
+                           (roslisp:ros-error (pr2-proj tf) "Somehow a virtual link with no parents in the bullet world was found in the urdf. This should never happen.")
+                           (return))
```
gaya-: Does the `return` exit from the whole function? Is that correct behavior? Should it not simply exit the virtual link assertion part?

cram_pr2/cram_pr2_projection/src/tf.lisp
```diff
@@ -43,43 +44,81 @@
...
-    ;; execute the TF update callbacks
-    (cl-tf:execute-changed-callbacks transformer))))
+      ;; tell the tf transformer about the virtual fixed joints
+      (let* ((bullet-links (btr:link-names ?robot-instance))
+             (urdf-links (cl-urdf:links (btr:urdf ?robot-instance)))
+             ;; get the names of the links, which are only present in the urdf
+             (urdf-link-names (remove-if (lambda (x) (member x bullet-links))
+                                         (loop for name being the hash-keys in urdf-links
+                                               collecting name))))
+        (let ((prev-link-count))
```
gaya-: The second bracket is not necessary here, it is indeed a bit confusing. It should either be something like: 
```lisp
(let ((prev-link-count 0)) ...)
```
or
```lisp
(let (prev-link-count) ...)
```

cram_pr2/cram_pr2_projection/src/tf.lisp
```diff
@@ -43,43 +44,81 @@
...
+          ;; while there are still links left to be asserted in tf
+          (loop while (> (length urdf-link-names) 0)
+                do ;; if no links were asserted last loop, we are stuck and have to return
+                   (if (eq prev-link-count (length urdf-link-names))
+                       (progn
+                           (roslisp:ros-error (pr2-proj tf) "Somehow a virtual link with no parents in the bullet world was found in the urdf. This should never happen.")
+                           (return))
+                       (setf prev-link-count (length urdf-link-names)))
+                   (let ((next-round (list)))
```
gaya-:
```lisp
(let ((next-round (list)))
```
is the same as
```lisp
(let (next-round)
```

cram_pr2/cram_pr2_projection/src/tf.lisp
gaya-: Also please use `when` for `if`-s, which don't have an `else` statement.

gaya-:
So I tested the PR, seem to work quite well.
Please commit the minor changes that I mention in the comments, then I will merge this PR.

cram_pr2/cram_pr2_projection/src/tf.lisp
cpollok: As I tested it `return` exits the loop macro. The HyperSpec states it returns from the block named `nil`, which is the name of implicitly created blocks like from `do`. I can't test this right now, so this would have to wait until next week. But I think it works as I intended it to.

cpollok:
@gaya- The last commit adresses all of your comments except for the `return` thing. I can't test this today and the changes were made using a normal text editor, so if you want to merge you should run it one more time just to be sure.

cram_pr2/cram_pr2_projection/src/tf.lisp
```diff
@@ -86,24 +86,26 @@
...
                      ;; try to assert a transform from a parent in the bullet world to the virtual links
                      (loop for link-name in urdf-link-names
                            do (let* ((link-joint (cl-urdf:from-joint (gethash link-name urdf-links)))
                                      (parent-name (cl-urdf:name (cl-urdf:parent link-joint))))
                                 (when (and (not (eq (cl-urdf:joint-type link-joint) :FIXED))
                                            display-warnings)
                                   (roslisp:ros-warn (pr2-proj tf)
-                                                    "Joint of ~a is ~a. Only links on FIXED joints should be asserted as virtual links."
+                                                    "Joint of ~a is ~a. Only links on FIXED joints should ~
+													be asserted as virtual links."
```
gaya-: It might be just the GitHub visualization but the line is still way too long (> 80 characters). If you remove more whitespace before the second line it will get shorter.

gaya-:
@cpollok I'll wait for you to test it and finish the rest.

cpollok:
Finally tested and (hopefully) finished.

gaya-:
@cpollok Thanks for the PR!

PR #47, opened 2018-05-07
Add another key to reference a location

artnie:
If two keys for location designators are not enough, this adds the :located-at key. I used it for more complex and less compact action designators.

### DIFF
--- a/cram_common/cram_designator_specification/src/specs.lisp
+++ b/cram_common/cram_designator_specification/src/specs.lisp
@@ -97,7 +97,7 @@
 
   (<- (property ?designator (?location-key ?location))
     (lisp-pred typep ?designator desig:action-designator)
-    (member ?location-key (:target :location))
+    (member ?location-key (:target :location :located-at))
     (property-member (?location-key ?location) ?designator)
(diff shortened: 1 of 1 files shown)


artnie:
The `(located-at ...)` keyword is only used in the context of transporting something from a container:
```
(desig:an action
    (type transporting-from-container)
    (object ?object-to-fetch)
    (located-at ?container-location)
    (location ?fetching-location)
    (target ?delivering-location))
```
It is located in the `cram_plan_transformation` repo, [here](https://github.com/artnie/cram_plan_transformation/blob/2522d819fc3d215e6f28d24236c509fa8ff5eab2/plan_transformation/lisp/top-level-plans.lisp#L197).

gaya-:
I will put this PR on hold for now, until the transporting from container plan is ready. I suspect that ``located-at`` will be automatically inferred, so there will be no need to explicitly specify it in the designator. Will keep the PR here until we know for sure.

gaya-:
``?fetching-location``is now used to infer ``?container-location``, so ``located-at`` is no longer necessary.

PR #49, opened 2018-05-07
Add attach items in bullet world.

artnie:
Using a carrier to transport multiple objects requires this implementation. You can attach one item to another, to make it move when the carrier moves.


gaya-:
Hmm, this PR turned into a big mess somehow... It would be nice if all the commits that don't belong to the PR would be gone. It might also be that some of this is already merged.

artnie:
Since this PR got out of hand I provide a fixed PR in #71. This PR can be closed when the other one has been accounted for.

PR #50, opened 2018-05-07
Add disable collisions parameter

artnie:
For now, it is not possible to move the gripper inside a mesh without it being recognized as a collision, even if the mesh is hollow and its inside basically accessible. This is due to the bounding box of meshes. As a hack, the alteration in this code makes it possible to temporarily disable collision detection with the kitchen.

### DIFF
--- a/cram_pr2/cram_pr2_projection_reasoning/src/check-collisions.lisp
+++ b/cram_pr2/cram_pr2_projection_reasoning/src/check-collisions.lisp
@@ -34,6 +34,8 @@
   "in seconds, sleeps after each movement during reasoning")
 (defparameter *debug-long-sleep-duration* 0.0
   "in seconds, sleeps to show colliding configurations")
+(defparameter *allow-kitchen-collision* nil
+  "set to T if collision with kitchen should be ignored. for reaching into containers.")
 
 (defun check-navigating-collisions (navigation-location-desig &optional (samples-to-try 30))
   (declare (type desig:location-designator navigation-location-desig))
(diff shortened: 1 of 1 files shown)


gaya-:
This will be very useful for the coming environment manipulation feature. We should look at this as soon as opening and closing drawers works well. Then we will look into getting objects out of drawers and that's where this PR will come handy, to disable collisions with drawers. Might require some minor changes though, so I'm not merging it yet, will look closer next week. @cpollok

gaya-:
(Mentioning @cpollok because it might be interesting for him.)

gaya-:
Clean solution suggested originally by Arthur.
Make Bullet understand concave collision shapes.
ToDo list:

1. look at sample .obj file
2. learn how obj files are defined (how are groups written)
3. load obj file with assimp: 
(cram-physics-utils::ai-import-file
       "/home/gaya/workspace/ros_lisp/src/cram/cram_pr2/cram_pr2_pick_place_demo/resource/bowl.stl"
       (cram-physics-utils::convert-post-process-enums-to-bitfield
        `(:join-identical-vertices
          :gen-smooth-normals
          :fix-infacing-normals
          :triangulate)))
4. look at number of meshes inside:
(cram-physics-utils::get-meshes *  :post-process-vertex-index #'identity)
inspect to see how big the vector is (or use (length *))
5. if assimp understands obj files properly, create compound-3d-model struct. Add a flag to LOAD-3D-MODEL which decide if a compound model should be created or not. Flag called USE-COMPOUND-MESH which is always NIL per default. If it is T, create compound-3d-model.
6. in cram_bullet_reasoning/src/robot-model.lisp use the new flag in LOAD-MESH
7. in the same file, in URDF-MAKE-COLLISION-SHAPE use a compound-mesh-shape instead of convex-hull-mesh-shape. For that, create a constructor for compound mesh shapes that goes over the list of meshes in the loaded model and calls ADD-CHILD-SHAPE. See btr/src/items.lisp MAKE-OCTAGON-PRISM-SHAPE as example. To see the pose of the compound shape children, look for example in phys_utils/src/assimp-model-loader.lisp BUILD-MODEL-MESH (let ((NODE-TRANSFORMATION...)))

gaya-:
Clean solution is finished. Solved in https://github.com/cram2/cram/pull/66. Thanks a lot @artnie

PR #51, opened 2018-05-07
Add variable for visibility threshold

artnie:
Object detection required 90% of the object to be visible, or else it is not detected. When stacking objects, the bottommost may get occluded easily. This implementation lets the programmer tweak the allowed level of occlusion.


gaya-:
Nice feature, thanks. Could you please export the symbol for the new variable you introduced, such that it can be used with ``:`` instead of ``::``?

artnie:
variable is exported

gaya-:
Merged into cram7
055e3308cacb6e344f83cb883f90e75558f8ef83

gaya-:
You commits are also in cram7, of course.

PR #52, opened 2018-05-07
Add ?retract-arms to fetch and deliver plans to prevent park-arms

artnie:
Sometimes the robot should not be allowed to retract his arms, when transporting an object (e.g. a loaded tray). Therefore setting ?retract-arms to NIL prevents performing park-arms, after lifting an object. This parameter is optional and T by default.


gaya-:
This doesn't seem like a very clean solution. It is indeed a problem if the robot parks him arms and flips the tray doing so, but completely avoiding to retract the arms is not a solution. There should be a good (possibly different from current default) carry pose for trays as well. Otherwise the robot will hit the environment with the tray when navigating around.

gaya-:
Should be implemented in a cleaner way. Someday...

PR #57, opened 2018-05-25
Environment Manipulation Fixes

cpollok:
Defines the 2nd-pregrasp-pose for containers and also includes the calculation of pose and transform of containers in the action designator reference. So the designators don't have to pre-calculate a pose, to be able to be referenced.

### DIFF
--- a/cram_pr2/cram_pr2_environment_manipulation/src/action-designators.lisp
+++ b/cram_pr2/cram_pr2_environment_manipulation/src/action-designators.lisp
@@ -29,6 +29,17 @@
 
 (in-package :pr2-em)
 
+(defun get-container-pose-and-transform (name)
+  (let* ((name-rosified (roslisp-utilities:rosify-underscores-lisp-name name))
+         (urdf-pose (get-urdf-link-pose name-rosified))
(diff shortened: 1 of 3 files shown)


cram_pr2/cram_pr2_environment_manipulation/src/action-designators.lisp
```diff
@@ -29,6 +29,17 @@
 
 (in-package :pr2-em)
 
+(defun get-container-pose-and-transform (name)
+  (let* ((name-rosified (roslisp-utilities:rosify-underscores-lisp-name name))
+         (urdf-pose (get-urdf-link-pose name-rosified))
+         (pose (cl-tf:transform-pose-stamped cram-tf:*transformer*
```
gaya-: Please use ``cram-tf:ensure-pose-in-frame`` instead of ``cl-tf:transform-pose-stamped``. And maybe use zero time.

gaya-:
Ok, minor thing, did it myself.

Thanks.

It works fine. Closing already closed containers is weird, so will have to keep that in mind. But for now will do :)

PR #64, opened 2018-06-11
Add dependency to knowrob in bullet tutorial.

artnie:


gaya-:
Please also add it to the package.xml. In general, all dependencies should be specified in both *.asd and package.xml.

PR #68, opened 2018-06-22
First steps for adding revolute joints into environment manipulation

cpollok:
This only includes a change to get-object-grasping-poses, which tries to find the most specific object-type when choosing a method for the generic grasp functions. This is based on the hierarchy in the knowrob_pnp/type-hierarchy.lisp file.

At some point I must have accidently merged my semi-private demo-branch into this. But you can safely jsut delete the cram_pr2_environment_manipulation_demo package. I don't think it belongs into the official cram stack in its current state. I'm really sorry for that. Will have to be more careful in the future.


gaya-:
Unfortunately doesn't work with the other items, such as bowls or cups.

gaya-:
@cpollok Oh, I deleted the cram7 branch as it got merged into master, so this PR was automatically closed. I'll see if I can cherry-pick manually. Otherwise we reopen it when you're back

Issue #54, opened 2018-05-22
Asserting joint states in bullet world does not respect joint limits


gaya-:
Use the URDF to read the joint limits and always crop to the limits.

gaya-:
Solved in cb4e8e65d47db46bb8e1f99fdc402c4b0a008033 and f1935710c95b95785189be17b4fa7d53430a4c39

PR #70, opened 2018-07-04
[bullet] Restrict joint states to limits. Solves issue #54

artnie:
#54 Solved. Requested value for a joint too high? -> take upper limit. Too low? -> take lower. Everything within is fine. Requires kitchen to be loaded as URDF (comes with cram7).

### DIFF
--- a/cram_3d_world/cram_bullet_reasoning/src/robot-model.lisp
+++ b/cram_3d_world/cram_bullet_reasoning/src/robot-model.lisp
@@ -453,7 +453,7 @@ current joint states"
            (parent (cl-urdf:parent joint))
            (parent-pose (find-parent-pose obj name))
            (limits (cl-urdf:limits joint)))
-      (when limits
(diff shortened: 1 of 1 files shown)


gaya-:
There's a bug.

gaya-:
It's probably about the infinite joints. Maybe need to normalize the angle or something.
Please test on a pick and place scenario or something like that.

artnie:
Some joints have upper and lower state limits of '0' equally. Those cases should be ignored now. They appear at the joints for the wheels and the wrist rolls.

gaya-:
@artnie Checking for infinite joints is hacky. You check if lower and upper limits are the same. In reality, this can actually happen, for some joints that are fixed in place and cannot move. Please use the joint type for ignoring infinite joints. They are actually called "continuous" joints, so if the type of the joint is equal to that, you can skip joint limit checks.

gaya-:
@artnie Ping. The last comment is fixed very easily.

gaya-:
Manually cherry-picked into boxy branch.

PR #77, opened 2018-11-23
Concave collision shapes and moving supported objects

mpomarlan:
Contents of the pull request:

- a few mesh files, in particular new daes for collision shapes. The bowl mesh is different, but has same size (diameter, height) as the original.
- code to optionally load compound collision shapes. This allows a collision shape to be made up of more bits, where each bit is convex (which helps collision checking) while the full collision shape may be concave. Also has a mini-test for these compound objects, the pile of plates scenario.
- code to move supported objects when the object supporting them moves because of a link attachment.

On the last point, I do not attach supported objects to their support, nor do I attach them to the link. This makes it possible to place a cup on a plate, then grab-lift the plate with the left arm, then grab the cup from the plate with the right arm, and no detachment of the cup needed from the left arm.

Support relation is computed via the prolog supported-by predicate. This unfortunately is not ideal, and will not fix your "fork in drawer should move with drawer" problem. As written, supported-by tests that two objects are in contact, and the center of the bounding box of one is above the center of the bounding box for the other. This will work in many cases (stacking plates, putting items on a tray), but not always (small item in large container, e.g. fork in drawer).

Note that when computing who should move, indirect support is computed. If the tray is attached to the arm, and the tray supports the plate, and the plate supports the cup, then when the arm move so will the tray, plate, and cup.


gaya-:
Hey @mpomarlan, please don't add new commits to existing pull requests, mixing different features together makes merging pull requests difficult. This might have been unintentional. To avoid this, it is usually nice to create a new branch on one's own fork and push the specific commits for the PR there, then switch back to master and develop there.

mpomarlan:
Thanks for the heads-up. In this particular case it was intentional, as I thought this new feature (getting a ref-obj size and padding out of a location desig, assuming said location desig has a for property) is small enough and of general interest enough.

But yeah, I can see that it will get annoying if I keep adding new commits here, so for now I'll stop :) Let's look at these features first then.

gaya-:
You can always create new PRs with new features, just on a different branch.

gaya-:
The functionality of this PR is more or less already implemented:

* a few mesh files, in particular new daes for collision shapes. The bowl mesh is different, but has same size (diameter, height) as the original.

-> we have convex decompositions of multiple concave objects, such as bowls, cups, plates etc

* code to optionally load compound collision shapes. This allows a collision shape to be made up of more bits, where each bit is convex (which helps collision checking) while the full collision shape may be concave. Also has a mini-test for these compound objects, the pile of plates scenario.

-> this is also implemented now, only we don't use an additional visual mesh, we work with collision meshes so they don't really have to look good.

* code to move supported objects when the object supporting them moves because of a link attachment.

-> we have an elaborate mechanism for attachments, and placing a cup on a tray and then moving the tray such that the cup follows and picking up the cup such that the tray doesn't follow is implemented via loose attachments.

So I'm closing this PR, as there is not much new in it left anymore and some bits were already taken over by cherry-picking etc.

PR #100, opened 2019-03-19
Added new failure handling strategy

amar-fayaz:


cram_common/cram_common_failures/src/failure-handling-strategies.lisp
```diff
@@ -0,0 +1,94 @@
...
+             (roslisp:ros-warn ,name "No samples left ~%")))))
+       (roslisp:ros-warn ,name "No retries left.~%")
+       (if ,rethrow-failure
+           (cpl:fail ,rethrow-failure))))
+
+(defmacro retry-with-list-solutions (iterator-list
+                                     retries
+                                     (&key
+                                        error
+                                        reset-designators
```
gaya-: Please remove reset-designators in the list macro.
We are iterating through lists, so there shouldn't be designators depending on the list.
If we end up needing it in the future, we will add it back.

cram_common/cram_common_failures/src/package.lisp
```diff
@@ -77,4 +77,6 @@
    ;; torso
    #:torso-low-level-failure
    #:torso-goal-unreachable
-   #:torso-goal-not-reached))
+   #:torso-goal-not-reached
+   ;; failure-handling-strategies
+   #:retry-with-designator-solutions))
```
gaya-: Please also export the list macro name.

cram_common/cram_common_failures/src/failure-handling-strategies.lisp
```diff
@@ -0,0 +1,91 @@
...
+;;; CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+;;; ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+;;; POSSIBILITY OF SUCH DAMAGE.
+
+(in-package :common-fail)
+
+(defmacro retry-with-designator-solutions (iterator-desig
+                                           retries
+                                           (&key
+                                              error
```
gaya-: I would also rename ``error`` into ``error-object`` and switch places of  ``reset-designators`` and ``name`` in the key list, the key list doesn't have ordering but it would look nicer if ``name`` and ``error-object`` are together.
And, finally, ``name`` as a variable name is a bit confusing. It is only used for warning prints. So how about ``warning-namespace``? Or something different, just more descriptive than ``name``. Otherwise it looks good.

cram_common/cram_common_failures/src/failure-handling-strategies.lisp
```diff
@@ -81,11 +81,56 @@ after each iteration of the retry."
...
+               (loop for designator in ,reset-designators
+                     do (desig:reset designator))
+               ,@body
+               (cpl:retry)
+             (roslisp:ros-warn ,warning-namespace "No samples left ~%")))))
+       (roslisp:ros-warn ,warning-namespace "No retries left.~%")
+       (if ,rethrow-failure
+           (cpl:fail ,rethrow-failure))))
+
+(defun next-different-location-solution (designator &optional (threshold 0.05))
```
gaya-: Could you please put ``next-different-location-solution`` above the macro that uses it?
SBCL can deal with things like that without throwing warnings but just to be sure.

cram_common/cram_common_failures/package.xml
```diff
@@ -17,7 +17,13 @@
...
 
+  <build_depend>cram_designators</build_depend>
   <build_depend>cram_language</build_depend>
+  <build_depend>cram_tf</build_depend>
+  <build_depend>roslisp</build_depend>
 
+  <run_depend>cram_designators</run_depend>
   <run_depend>cram_language</run_depend>
+  <run_depend>cram_tf</run_depend>
+  <run_depend>roslisp</run_depend>
```
gaya-: Now that you're at it anyway, could you please change the package.xml to version v2?
Such that we don't have any ``build_depend`` and ``run_depends`` but just ``depend``?
See, for example, https://github.com/cram2/cram/blob/master/cram_common/cram_object_interfaces/package.xml

PR #101, opened 2019-04-01
[cram-urdf-projection] new pkg to combine all similar code from the robots in cram.

Vanessa-rin:
This is still work in progress. I am combining similar code in to the new package.


cram_urdf_projection/.gitignore
```diff
@@ -0,0 +1,5 @@
+*~
```
gaya-: Currently, this package is in the root of CRAM. We only add new directories to the root if it's really necessary. This package should rather be in cram_3d_world.

cram_urdf_projection/cram_urdf_projection/src/giskard.lisp
```diff
@@ -0,0 +1,404 @@
+;;;
```
gaya-: This does not need to be here, please delete.

cram_urdf_projection/cram_urdf_projection/package.xml
```diff
@@ -0,0 +1,38 @@
...
+  <depend>cram_common_designators</depend>
+  <depend>cram_boxy_designators</depend>
+  <depend>cram_common_failures</depend>
+  <depend>cram_process_modules</depend>
+  <depend>roslisp_utilities</depend>
+  <depend>alexandria</depend>
+  <depend>cram_semantic_map</depend>
+  <depend>cram_bullet_reasoning_belief_state</depend>
+  <depend>cram_simple_actionlib_client</depend>
+  <!-- <depend>giskard_msgs</depend> -->
```
gaya-: Please delete this comment.

cram_urdf_projection/cram_urdf_projection/cram-urdf-projection.asd
```diff
@@ -0,0 +1,66 @@
...
+               cram-bullet-reasoning-belief-state ; for special projection variable definition
+               cram-simple-actionlib-client ; for communicating with giskard
+               moveit_msgs-msg ; for IK queries
+               moveit_msgs-srv)
+  :components
+  ((:module "src"
+    :components
+    ((:file "package")
+     (:file "tf" :depends-on ("package"))
+     ;; (:file "giskard" :depends-on ("package"))
```
gaya-: Please delete this comment.

cram_urdf_projection/cram_urdf_projection/src/ik.lisp
```diff
@@ -0,0 +1,212 @@
...
+              :seed-state seed-state
+              :test-angle test-angle
+              :torso-angle torso-angle
+              :torso-lower-limit torso-lower-limit
+              :torso-upper-limit torso-upper-limit))
+        (roslisp:set-debug-level NIL old-debug-lvl)))))
+
+
+
+#+commentedoutfornow
```
gaya-: This can be thrown away. For now I mean the commented out stuff.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
```diff
@@ -0,0 +1,583 @@
...
+          (btr:bullet-world ?w)
+          (btr:assert ?w (btr:joint-state ?robot ,joint-states))))))
+
+(defun look-at (pose configuration)
+  (let* ((neck-joints-amount (- (length (third (car (prolog:prolog
+                                                     `(and (cram-robot-interfaces:robot ?robot)
+                                                           (btr:bullet-world ?w)
+                                                           (cram-robot-interfaces:robot-pan-tilt-joints
+                                                            ?robot . ?joint-names))))))1))) 
+                             (if  (and configuration (> neck-joints-amount 2))
```
gaya-: The indentation is misaligned.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
```diff
@@ -0,0 +1,583 @@
...
+    `(and (cram-robot-interfaces:robot ?robot)
+          (btr:bullet-world ?w)
+          (btr:assert ?w (btr:joint-state ?robot ,joint-states))))))
+
+(defun look-at (pose configuration)
+  (let* ((neck-joints-amount (- (length (third (car (prolog:prolog
+                                                     `(and (cram-robot-interfaces:robot ?robot)
+                                                           (btr:bullet-world ?w)
+                                                           (cram-robot-interfaces:robot-pan-tilt-joints
+                                                            ?robot . ?joint-names))))))1)))
```
gaya-: This is very hard to read. Please put the 1 on the next line.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
```diff
@@ -0,0 +1,583 @@
...
+(defun look-at-joint-states (joint-states)
+  (declare (type list joint-states))
+  (assert
+   (prolog:prolog
+    `(and (cram-robot-interfaces:robot ?robot)
+          (btr:bullet-world ?w)
+          (btr:assert ?w (btr:joint-state ?robot ,joint-states))))))
+
+(defun look-at (pose configuration)
+  (let* ((neck-joints-amount (- (length (third (car (prolog:prolog
```
gaya-: If there is only one variable, there is no need for ``let*``, ``let`` should suffice.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
```diff
@@ -0,0 +1,583 @@
...
+
+(defun look-at-joint-states (joint-states)
+  (declare (type list joint-states))
+  (assert
+   (prolog:prolog
+    `(and (cram-robot-interfaces:robot ?robot)
+          (btr:bullet-world ?w)
+          (btr:assert ?w (btr:joint-state ?robot ,joint-states))))))
+
+(defun look-at (pose configuration)
```
gaya-: It would be more consistent if ``look-at`` came after all the 3 variations of look, i.e. after ``look-at-pose-stamped``, please change the places of ``look-at`` and ``look-at-pose-stamped``.

cram_urdf_projection/cram_urdf_projection/src/process-modules.lisp
```diff
@@ -0,0 +1,104 @@
...
+      (cram-common-designators:move-base (drive argument)))))
+
+;;;;;;;;;;;;;;;;; TORSO ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(cpm:def-process-module urdf-proj-torso (motion-designator)
+  (destructuring-bind (command argument) (desig:reference motion-designator)
+    (ecase command
+      (cram-common-designators:move-torso (move-torso argument)))))
+
+;;;;;;;;;;;;;;;;; PTU ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
gaya-: This should probably be renamed into NECK, so ``urdf-proj-neck``

cram_urdf_projection/cram_urdf_projection/src/setup-urdf.lisp
```diff
@@ -0,0 +1,97 @@
+;;;
```
gaya-: This file is very robot-specific and is a setup thing. Please move this file into urdf_bringup package.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
```diff
@@ -372,9 +388,9 @@
     (set-configuration :right right-configuration)))
 
 ;;; cartesian movement
-(defparameter *gripper-length* 0.2 "PR2's gripper length in meters, for calculating TCP -> EE")
+
 (defparameter *tcp-pose-in-ee* (cl-transforms:make-pose
-                                (cl-transforms:make-3d-vector (- *gripper-length*) 0 0)
+                                (cl-transforms:make-3d-vector (- (get-gripper-length)) 0 0)
```
gaya-: This will not work if the TCP is not on the X axis of EE, but on Y or Z axis. Please use a full vector.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
```diff
@@ -339,6 +339,22 @@
...
+              (prolog:prolog
+               `(and (cram-robot-interfaces:robot ?robot)
+                     (cram-robot-interfaces:robot-tool-frame ?robot :left ?left-tool-frame)
+                     (cram-robot-interfaces:end-effector-link ?robot :left ?left-ee-frame)))))
+           (left-tcp-frame
+             (cut:var-value '?left-tool-frame frame-bindings))
+           (left-ee-frame
+             (cut:var-value '?left-ee-frame frame-bindings)))
+    (roslisp:with-fields (translation)
+        (cl-tf:lookup-transform cram-tf:*transformer* left-tcp-frame left-ee-frame)
```
gaya-: Please use ``cl-transforms-stamped`` for such things.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
```diff
@@ -339,6 +339,22 @@
...
+      (cl-tf:v-dist (cl-tf:make-3d-vector 0 0 0) translation))))
```
gaya-: Please use ``cl-transforms`` for such things.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
```diff
@@ -339,6 +339,22 @@
...
+
```
gaya-: But actually, I'm not sure that I like this solution with TF. What if there is no TF listener. In theory, the transform between the EE and TCP should be already given in the URDF, as the TCP is fixed onto EE, the joint that connects them is always fixed (unless the person making a URDF did something weird). Please look at all the 3 URDFs of the robots that you support and find the fixed joint that connects EE to TCP of all the robots. And write here what you find out. You can use RViz TF tree to search for this. If the chain between EE and TCP is not direct (it should be EE frame -> TCP frame and not EE frame -> some other frame -> and another frame -> TCP), write down the full chain. It should be that all of them are connected directly and you can find the gripper-offset from that connection.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
gaya-: This is definitely not going to work, `defparameter`` gets evaluated at compile time, when the TF listener is not initialized yet. Did you test this? Solving the problem without using TF whatsoever will solve this issue with ``defparameter`` as well.

cram_3d_world/cram_urdf_projection/cram_urdf_projection/src/process-modules.lisp
```diff
@@ -0,0 +1,104 @@
+;;;
+;;; Copyright (c) 2017, Gayane Kazhoyan <kazhoyan@cs.uni-bremen.de>
+;;;               2019, Vanessa Hassouna <hassouna@uni-bremen.de>
```
gaya-: Hm, I don't see any substantial contributions from you in this file, so I think it doesn't make sense to put your name here. I do see contributions from you in low-level.lisp though, maybe that's what you wanted to do in the first place.

cram_3d_world/cram_urdf_projection/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.~
```
gaya-: Let's do it like this:

cram_urdf_projection goes to cram_3d_world:

    cram_3d_world/cram_urdf_projection/package.xml

and cram_urdf_bringup goes to somewhere else :) It depends on robot URDFs etc, so definitely shouldn't be in cram_3d_world. For now, I'd say let's put it into cram_pr2 and later find a better place for it once more packages are robot-independent:

    cram_pr2/cram_urdf_bringup/package.xml

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
Vanessa-rin: So after quick research: For pr2 and hsrb the chain is: wrist_roll_link, hand_palm_link gripper_tool..
but for boxy i see *_arm_7_link -> right_arm_flange_link -> .... 11 more ... -> right_gripper_base_link -> right_gripper_tool_frame

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
Vanessa-rin: deleted this for now since we searching for a better solution

cram_urdf_projection/cram_urdf_projection/src/process-modules.lisp
Vanessa-rin: should i let   
`  (cpl:fail 'common-fail:ptu-goal-not-reached
                :description "Look action wanted to twist the neck"))))`  like that or should i also change the common-fail to neck?

gaya-: Well that would be nice if you renamed it but it might be used all over the place. So let's leave it for now. First finish this PR and then rename the failure stuff.

cram_urdf_projection/cram_urdf_projection/src/low-level.lisp
Vanessa-rin: new solution is up

PR #102, opened 2019-04-02
Implement manipulation of rotational joints

cpollok:
Adds support for rotational joints to the environment-manipulation support.
At least for those rotating around Z. For other axes (for example the oven) it does not work yet.

To just open the fridge (given the PR2 stands in a suitable position), you can use:
```
(pr2-proj:with-simulated-robot
        (exe:perform
         (an action
             (type opening)
             (arm right)
             (distance 0.3)
             (object
              (an object
                 (type container-revolute)
                 (urdf-name :iai_fridge_door)
                 (part-of kitchen))))))
```

To approach and open the fridge one could use the accessing plan like so:
```
(pr2-proj:with-simulated-robot
        (exe:perform
         (an action
             (type accessing)
             (arm right)
             (distance 0.4)
             (location
              (a location
                 (in
                  (an object
                      (type container-revolute)
                      (urdf-name :iai_fridge_door)
                      (part-of kitchen))))))))
```

Closing works analogue with closing and sealing.


cram_common/cram_manipulation_interfaces/src/trajectories.lisp
```diff
@@ -83,19 +83,19 @@
...
+                (cram-tf:multiply-transform-stampeds
+                 cram-tf:*robot-base-frame* gripper-tool-frame
+                 object-transform          ; bTo
+                 object-to-standard-gripper ; oTg'
+                 :result-as-pose-or-transform :transform))) ; bTo * oTg' = bTg'
+          (cram-tf:multiply-transform-stampeds ; bTg' * g'Tg = bTg
+           cram-tf:*robot-base-frame* gripper-tool-frame
+           base-to-standard-gripper-transform      ; bTg'
+           standard-to-particular-gripper-transform ; g'Tg
+           :result-as-pose-or-transform :pose))))))
```
gaya-: Please don't commit things like this. They create unnecessary merge conflicts just because your emacs has a different preference for indentation than mine.

cram_pr2/cram_pr2_environment_manipulation/src/trajectories.lisp
```diff
@@ -126,43 +183,114 @@
...
+                      grasp-pose :x-offset (+ opening-distance *drawer-handle-retract-offset*))))))
 
-(defmethod man-int:get-action-trajectory ((action-type (eql :opening))
-                                           arm
-                                           grasp
-                                           objects-acted-on
-                                           &key
-                                             opening-distance)
-  (make-trajectory action-type arm objects-acted-on opening-distance))
+(defun make-revolute-trajectory (object-transform arm action-type
```
gaya-: This returns the same trajectory for the opening and closing action, whereas they are actually supposed to be inverted. I do not see anywhere in the code the inversion. Have you tested your feature with a closing action?

gaya-: Ah, you invert the opening-distance, i see. Please ignore the above comment.

cram_common/cram_manipulation_interfaces/src/trajectories.lisp
gaya-: This relates more to common packages that multiple people might be working on at the same time. In your packages fill free to fix the indentation but in common packages try to refrain from committing indentation fixes.

PR #106, opened 2019-04-12
Adding support to place objects inside Containers

amar-fayaz:
1. Made costmap changes required to support objects inside containers using 'for' keyword for 'in' based 
   locations :- 
`(a location 
   (in (an object
           (type drawer)
           (urdf-name sink-area-upper-left-drawer-main)
           (part-of kitchen)))
   (for (an object 
            (type bowl))))`
2. Modified the 'transport' plan to support placing objects into inaccessible locations (containers)


gaya-:
.It's ``sink-area-left-upper-drawer-main`` not ``sink-area-upper-left-drawer-main``

PR #108, opened 2019-04-24
Attachments for items and generalization from methods attach-object, detach-object ...

tlipps:
The code from grasping.lisp and placing.lisp are now in assembly.lisp [here](https://github.com/tlipps/cram/commit/2b6ffe95ac35443b6d46431e64f11aad9eb17dce#diff-41). Only shown here as a link, because it was edited in in a merge commit.


gaya-:
Merged manually in 90329b4d47d2ac5ac34d2b25c81c1a1bd1e4fde5.

PR #109, opened 2019-04-26
Add helper function for calling generic functions with most specific object-type

cpollok:
Just what the title says. Adds a simple defun to trajectories.lisp that encapsules the call to find-most-specific-object-type-for-generic and the associate error handling.

### DIFF
--- a/cram_common/cram_manipulation_interfaces/src/gripper.lisp
+++ b/cram_common/cram_manipulation_interfaces/src/gripper.lisp
@@ -30,27 +30,7 @@
 (in-package :cram-manipulation-interfaces)
 
 (defmethod get-action-gripping-effort :heuristics 20 (object-type)
-  (let ((specific-type
-          (find-most-specific-object-type-for-generic
-           #'get-action-gripping-effort object-type)))
-    (if specific-type
-        (get-action-gripping-effort specific-type)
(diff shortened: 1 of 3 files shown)


cram_common/cram_manipulation_interfaces/src/gripper.lisp
```diff
@@ -30,27 +30,7 @@
...
-           #'get-action-gripping-effort object-type)))
-    (if specific-type
-        (get-action-gripping-effort specific-type)
-        (error "There is no applicable method for the generic function ~%~a~%~
-                with object-type ~a.~%To fix this either: ~
-                ~%- Add a method with (object-type (eql ~a)) as the first specializer or ~
-                ~%- Add ~a into the type hierarchy in the cram_object_knowledge package."
-               #'get-action-gripping-effort
-               object-type object-type object-type))))
+  (call-with-specific-type #'get-action-gripping-effort object-type))
```
gaya-: This introduces a dependency between gripper.lisp and trajectories.lisp. Either put this dependency into asd file or put ``call-with-specific-type`` somewhere else, maybe into ``manipulation-interfaces.lisp``.

cpollok: Oh yes. Sorry about that. Moved it to manipulation-interfaces.lisp.

PR #114, opened 2019-05-13
[br] robot-model also setf the joint of the mimics-joints 

Vanessa-rin:
When a joint is moved we search in the urdf for all the mimic joints that mimics this special joint. If there are any they also gets sets with the corresponding multiplier.


gaya-:
@Vanessa-rin you forgot the offset.

PR #118, opened 2019-05-27
[env-man] Change orientation in drawer costmap.

cpollok:
In this PR the orientation generator for opening drawers is changed. Instead of a direction opposite to the drawers opening trajectory, the robot is now rotated slightly towards the drawer.

This was introduced to give Boxy more room to work in. The PR2 seems to work with it just fine, too.


gaya-:
Awesome!

PR #120, opened 2019-06-17
Gravity-Fix: (Prev. Change attached objects collision flag to cf-kinematic-object.)

artnie:
Attaching an object to the robot has given the attached object the collision flag `:cf-static-object`, such that it is immune to simulation, collision and gravity. The navigating plan in `fetch-and-deliver-plans` includes collision checking before moving to the robot's destination, where it tries to fit the robot in the environment and resets the world if it doesn't fit. This re-spawns all rigid bodies in the world to their original place respectively. An object spawned with the `:cf-static-object` flag in its initial state stays static throughout its existence in the bullet world and can only be simulated by removing its rigid body from the bullet world, changing the parameters (essentially setting mass and removing the static flag) and re-adding the rigid-body to the bullet world. An object spawned as `dynamic` (default) can be set to static and back to dynamic again, without further ado.

While the robot is holding an object, that attached object has been flagged with `:cf-static-object` and was re-spawned due to the collision checking in the navigation plan. This new static rigid-body is never affected by simulation. Instead of using the `:cf-static-object` flag, this commit flags attached objects as `:cf-kinematic-object`, which is meant for objects like tools or... hey nice, pickable and placable stuff. Essentially, those things are manipulated by an actor in the physics scene. They collide with dynamic objects and make them fly around. Think about a baseball bat. Such objects are not affected by simulation while flagged as is, but the flag can be removed, such that the default flag (dynamic) will take action again. 

See the Bullet Physics Manual at the bottom of page 26:
http://www.cs.kent.edu/~ruttan/GameEngines/lectures/Bullet_User_Manual

### DIFF
--- a/cram_3d_world/cl_bullet/src/lisp/package.lisp
+++ b/cram_3d_world/cl_bullet/src/lisp/package.lisp
@@ -184,4 +184,5 @@
    get-state world-state
    restore-world-state
    restore-state
-   copy-world))
+   copy-world
+   #:respawn-rigid-body-with-flags))
--- a/cram_3d_world/cl_bullet/src/lisp/world-state.lisp
+++ b/cram_3d_world/cl_bullet/src/lisp/world-state.lisp
@@ -241,3 +241,42 @@ world."
   (let ((constraint (make-motor-constraint 'slider-constraint world constraint)))
     (add-constraint world constraint)
     constraint))
+
+(defun respawn-rigid-body-with-flags (body world new-collision-flags)
+  "Creates a new `rigid-body-state' from the existing, with the
+given `new-collision-flags' as collision-flags. Then removes the current `rigid-body'
+and uses the new rigid-body-state to respawn it. Since rigid-body-state
+instances are not meant to be changed, we need this cloning procedure."
+  (let ((body-state (find (name body)
+                          (bodies (get-state world))
+                          :key 'name)))
+    (when body-state
+      (with-slots (name
+                   mass
+                   pose
+                   force torque
+                   linear-velocity
+                   angular-velocity
+                   activation-state
+                   collision-flags
+                   collision-shape
+                   collision-group
+                   collision-mask) body-state
+        (let ((new-body-state
+                (make-instance 'rigid-body-state
+                               :name name
+                               :mass mass
+                               :pose pose
+                               :force force
+                               :torque torque
+                               :linear-velocity linear-velocity
+                               :angular-velocity angular-velocity
+                               :activation-state activation-state
+                               :collision-flags new-collision-flags
+                               :collision-shape collision-shape
+                               :collision-group collision-group
+                               :collision-mask collision-mask)))
+          (with-world-locked world
+            (remove-rigid-body world body)
+            (restore-state new-body-state world))))))
+        world)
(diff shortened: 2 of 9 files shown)


gaya-:
@artnie This creates some very weird behavior. When transporting a bowl, it lies on a side on the kitchen island. When transporting a breakfast cereal, it ends up floating in the air 1 or 2 cm above kitchen island. Please test it with pick and place demo properly.

artnie:
This other fix re-spawns the rigid-body of a btr-object, after the robot has retracted its arm. When the re-spawning is done right after detaching, the mesh of the object is still in collision with the gripper and jumps aside, therefore the re-spawning is postponed until the gripper is retracted in the placing-plan. It fires a `object-placed-event` which contains the objects name, which triggers a method in the event handler, calling a method of the object to re-spawn the rigid-body on bullets side.

gaya-:
@artnie Please come by tomorrow when you're working to my office to talk about this.

gaya-:
@amar-fayaz please review this PR.

gaya-:
I still don't see why you need this object-placed-event. It seems like a hack to me. In the real world, the object starts falling down when you release it from the gripper and not when you moved the hand away. Besides, what if the placing trajectory would not have a lifting pose, you cannot guarantee that. You said the problem was that the hand was messing up with the collision stuff, so the object would jump away when released from the gripper. Could you please make some (informal) tests with this and some screenshots of which objects jump away and which don't. The gripper is a "compound" shape, so it shouldn't create weird collision behaviors. So, e.g., a spoon should not jump away. Neither should a breakfast cereal or a milk. What could jump away is the bowl, which is a concave object with a convex collision shape. Please make before and after pictures with all these objects and we see if `object-placed-event` is really necessary.

artnie:
Yes, this is a hack for now and does resemble the real world behaviour, but I wanted to make the objects in the placing plan work properly again. 
First I implemented the respawning of the rigid body just when it was detached from the gripper. Especially the bowl has been pushed away like 10cm when the gripper was still near/in the bounding box, other objects just moved a bit sometimes. This is the main reason, why I delayed the respawning until after retracting the gripper. Unless all our meshes are spawned as a compound of convex meshes (when they are hollow) I'm afraid this will happen all the time and compromise the position of a placed object. I will still provide the screenshots so you can decide.

gaya-:
Ok, so we decided to have the event thrown directly after releasing the object, such that the object can move slightly to a side when the simulation is started. In the future, we will implement support for concave manipulation items, such that things don't just too much.
@artnie please write a short comment for @amar-fayaz when you're done with refactoring, so he can test the PR again.

gaya-:
@amar-fayaz please test this once again and then this can be merged.

amar-fayaz:
@gaya- Tested, and still works.

gaya-:
@artnie I like the direction in which you were thinking but the solution wasn't perfect.
This only solves the problem of when the static object is detached during object manipulation.
We are just lucky that in that situation the collision flags are NIL.
There could have been a situation when the ``new_collision_flags == cf_static_object`` instead of NIL and that would make it not being affected by gravity forever.
Or until it is lucky enough to be respawned with ``new_collision_flags == NIL``.
A cleaner solution would be to make sure that no object can stay static forever
unless it is explicitly spawned like that, independent of the context,
be it attaching / detaching or whatever else.
I want to close the PRs that have been open for too long, and we've been on this one for at least 3 months now, so I'm going to solve it myself this time, while I have the time to look into it. But thanks a lot for looking into this and getting into the root of the issue
and providing a solution. I took your solution and applied it on a much lower level.

gaya-:
Solved in c2f94f1.

PR #121, opened 2019-06-18
made spawn locations in demo-random relative to furniture

Tigul:


gaya-:
@Tigul Please write a test for this, where you move the sink area using ``(setf (btr:link-pose ...`` and the objects follow the sink-area everywhere, even if it is rotated and translated.

cram_pr2/cram_pr2_pick_place_demo/src/projection-demo.lisp
```diff
@@ -70,8 +76,37 @@
     (:spoon . ((-0.78 1.5 0.86) (0 0 0 1)))
     (:milk . ((-0.75 1.7 0.95) (0 0 0.7071 0.7071)))))
 
-
-(defun spawn-objects-on-sink-counter (&optional (spawning-poses *object-spawning-poses*))
+(defun make-poses-relative (&optional (spawning-poses *object-spawning-poses*))
+  (let* ((object-types '(:breakfast-cereal :cup :bowl :spoon :milk))
+         (map->surface (cl-transforms:pose->transform
```
gaya-: Please don't use this notation: `map->surface`, it can be misleading. A transformation from `map` frame to `surface` frame can be understood as, you take a coordinate in the map frame: `mapP` and multiply it with your transform `Tmap->surface` and get a point in the surface frame `surP = Tmap->sur * mapP`. But this transform does the opposite: you input a point in the surface frame and get as an output a point in map. So please use a different notation: I prefer `map-T-surface` for a transform and `map-P-surface` for a pose. The only problem with this is that our IDE doesn't like capital letters inside of variable names.

cram_pr2/cram_pr2_pick_place_demo/src/projection-demo.lisp
```diff
@@ -70,8 +76,37 @@
...
+                       (btr:pose
+                        (btr:rigid-body
+                         (btr:object btr::*current-bullet-world* :kitchen) :|KITCHEN.sink_area_surface|))))
+         
+         (surface->objects (mapcar (lambda (type)
+                                     (setf coordinates (car (cdr (assoc type spawning-poses))))
```
gaya-: This is super ugly. You are defining a global variable `coordinates` right here in the lambda, which you only use later in the lambda. If you need a variable, use a `let` block. If you need a global variable, use a `defparameter` or `defvar` and make sure you name it with `*` on each side: `*my-global-var*`.

cram_pr2/cram_pr2_pick_place_demo/src/projection-demo.lisp
```diff
@@ -70,8 +76,37 @@
...
+                                       (third coordinates))
+                                      (cl-transforms:make-identity-rotation)))
+                                   object-types))
+         
+         (map->objects (mapcar (lambda (surface->object)
+                            (cl-transforms:transform* map->surface surface->object))
+                          surface->objects)))
+
+    (mapcar (lambda (transform type)
+              (setf tr (cl-transforms:translation transform))
```
gaya-: Same here, don't just put `setf` in a random place, where what you actually want is a `let` block.

Tigul:
I did the requested changes and added a few tests for the method

gaya-:
@Tigul these revert commits mess up the Pull Requests. Because of that now the other PR with kitchen variations has also been merged. Please don't commit revert commits, use `git reset --hard` very carefully instead.

PR #122, opened 2019-06-19
[fd-plans]Supporting 'dropping' action designator 

amar-fayaz:
_None of the keywords and names in this plan are confirmed and I don't expect it to be used in the final code. Will look for Gaya's suggestions on what to rename these keywords to._

This action primarily disables the stability check in the end of the deliver plan.
The deliver plan otherwise is completely the same and is reused for the dropping action
Additionally, the transport plan has been modified to also include a mode transport-and-drop(name to be changed to something appropriate) which will use the action 'dropping' instead of 'delivering'.

### DIFF
--- a/cram_3d_world/cram_fetch_deliver_plans/src/fetch-and-deliver-designators.lisp
+++ b/cram_3d_world/cram_fetch_deliver_plans/src/fetch-and-deliver-designators.lisp
@@ -54,9 +54,17 @@ the `look-pose-stamped'."
    (desig:reference target-designator)
    (cram-tf:robot-current-pose)))
 
-
-
-
+(defun replace-tag-in-location-designator (loc-designator old-tag new-tag)
+  (let ((?loc-designator-copy (desig:copy-designator loc-designator))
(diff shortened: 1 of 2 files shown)


amar-fayaz:
@gaya- : I have added the workaround you mentioned. The objects are transported below the floor after dropping action.

gaya-:
Waiting for the gravity fix so we can get rid of the Z offset hack for falling objects.

gaya-:
Checking for pose stability in the plan is hacky. I think the placing location designator should have `(type stable)` inside, which would then ensure somehow that stability check is done. Putting this on hold until a cleaner implementation is there. The Z offset hack can also now be removed.

gaya-:
@amar-fayaz oh you deleted the branch and automatically closed this PR :(

amar-fayaz:
Yeah I had too many stale branches, and accidentally ended up deleting this during cleanup. Restored.

gaya-:
Reimplementing in a cleaner way.

PR #123, opened 2019-06-21
README.md for level

amar-fayaz:
Added the README.md file which only contains help for 'level' keyword as of now.
Will add help for other keywords in a later changeset tracked by [Trello Card](https://trello.com/c/bFE71eKc/88-improve-location-costmap-tutorial-with-more-in-depth-infos)


amar-fayaz:
Also added info for other spatial relations as well.

PR #125, opened 2019-06-25
[env-man] Fixes for distance revision

cpollok:
Fixes the following bugs:
- Sealing action to distance 0.0 didn't work with goals. (Was closed to just 0.001 because of trying to avoid overshooting the limits.)
- Manpiulation of environment with distance being the same as the current joint-state raised an error. (No trajectory is calculated because the manipulation distance is too short. Just had to make sure there is no operation done on the nil result.)
- Rounding of angles before setting joint-state in event-handler just rounded up. That lead to problems when approaching the upper limit of joints.

### DIFF
--- a/cram_3d_world/cram_bullet_reasoning_belief_state/src/event-handlers.lisp
+++ b/cram_3d_world/cram_bullet_reasoning_belief_state/src/event-handlers.lisp
@@ -146,6 +146,7 @@ If there is no other method with 1 as qualifier, this method will be executed al
 
 
 (defun move-joint-by-event (event open-or-close)
+  (print event)
   (let* ((joint-name (cpoe:environment-event-joint-name event))
          (object (cpoe:environment-event-object event))
          (distance (cpoe:environment-event-distance event))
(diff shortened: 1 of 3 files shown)


cram_3d_world/cram_bullet_reasoning_belief_state/src/event-handlers.lisp
```diff
@@ -146,6 +146,7 @@ If there is no other method with 1 as qualifier, this method will be executed al
 
 
 (defun move-joint-by-event (event open-or-close)
+  (print event)
```
gaya-: What's that ;)

cram_3d_world/cram_bullet_reasoning_belief_state/src/event-handlers.lisp
```diff
@@ -158,16 +159,22 @@ If there is no other method with 1 as qualifier, this method will be executed al
...
+               new-joint-angle
+               0.0001)
+              10000)))
     (btr:set-robot-state-from-joints
      `((,joint-name
         ,new-joint-angle-rounded))
      object)))
 
 (defmethod cram-occasions-events:on-event open-container ((event cpoe:container-opening-event))
+  (print event)
```
gaya-: And that.

cram_3d_world/cram_bullet_reasoning_belief_state/src/event-handlers.lisp
cpollok: Whoops. Fixed.

cram_3d_world/cram_bullet_reasoning_belief_state/src/event-handlers.lisp
cpollok: Whoopsie. Fixed.

PR #126, opened 2019-06-26
Holes in global costmap after mapping a costmap rotated on it

tlipps:
In this PR the holes in the global costmap are fixed. The holes originated from the rotated input costmaps (learned from the cram_learning_framwork) which were mapped with the standard rotation matrix and landed sometimes in adjacent fields and created so holes in the global costmap. 
The problem was solved by mapping not only the point alone but the middle of the point too. This solution seemed to work pretty well and can be tested for the riskiest angles with the following code snippet: 
```
(loop for i from 2 to 4 do
          (let ((cm
                    (make-instance
                        'location-costmap:location-costmap
                      :resolution 0.02
                      :origin-x -6
                      :origin-y -6
                      :height 12.0
                      :width 12.0)))
              (costmap:register-cost-function 
               cm
               (costmap:make-matrix-cost-function 
                0.0 0.0 1.0
                (make-array '(4 4)
                            :element-type 'double-float
                            :initial-element 0.01d0)
                (/ pi i))
               :hello)
            (sleep 0.2)
            (costmap:costmap-samples cm)))
```


gaya-:
@tlipps You messed up your costmaps branch. Please do a git reset --hard HEAD^ (or HEAD^^) to get back to previous commit before the merge. That will completely remove the old commits, so store them somewhere. This PR should just have the one commit to fix the holes in rotated costmaps.

gaya-:
So the easiest way to do this is do a "git branch new-branch" from your current state of the "costmaps" branch. That makes a complete copy of "costmaps" into a new branch called "new-branch". Then do the "git reset --hard HEAD^^" to revert the merge. Then push the new branch to your fork as a new branch and change the PR to use the new branch instead of "costmaps".

gaya-:
Just in case, backup your cram directory to not lose any data.

tlipps:
Yes you are right. I could not change the source branch, so I simply created an new PR https://github.com/cram2/cram/pull/131.

PR #127, opened 2019-06-26
Fix door costmap

cpollok:
Fixes the door costmap so that it's not the same for left and right arms anymore.
The costmap now gets cut in front of the first 80% of the door's opening arc, when trying to open it with an opposing arm.
For example: The door of the fridge in the IAI kitchen opens to the right. So when trying to open it with the left arm, much of the costmap on the left side gets cut. That is to ensure, that the robot stands more to the right.
In general this should lead to robots standing in more benefical locations for opening containers with rotational joints.

PS.: This branch was forked from the cpollok-env-man-distance-revision branch to be able to use the features in there for smore smooth development. When that branch's PR is merged this PR should point to the branch it got merged into.


gaya-:
Merged into boxy-melodic.

PR #128, opened 2019-07-09
Perception noise

amar-fayaz:
[Trello](https://trello.com/c/Rjeuwy3P/125-implement-reasoning-on-world-state-for-updating-it-intelligently)
1. Added a method to simulate perception noise
2. Added a check to see if the perception noise caused an object instability after simulation
3. Added a method to attempt to correct the perception noise by moving the object a little bit to the direction opposite to the deviation in the object position.

Usage:
To simulate the noise in projection, set _btr-belief::*simulated-perception-noise-factor*_ to a positive value. Eg: 0.1 means the object can be moved around at most a distance of 0.1 meters on a randomly chosen axis. The default of this parameter is 0.0 which means no noise will be simulated in projection.

The parameter _btr-belief::*perception-instability-threshold* determines the threshold to activate the correction. Default is 0.1 meters

Similarly, another new parameter is _btr-belief::*perception-noise-correction-factor*_ which will determine the amount of distance the correction will be applied to. The default value of this is 0.04. Eg: if an object falls through a surface in the z-direction due to perception noise, the correction method will retry by spawning it +0.04 meters from the initial object position in the z axis. 

The correction method is activated if the perceived pose and the final pose of the object is greater than 1 meter. The correction method is only tried once and then it raises an error.


amar-fayaz:
Added unit tests for some corner cases

cram_3d_world/cram_bullet_reasoning_belief_state/cram-bullet-reasoning-belief-state.asd
```diff
@@ -38,6 +38,7 @@
                cram-common-designators ; for world-state-detecting motion
                cram-common-failures ; for world-state-detecting motion
                cram-bullet-reasoning
+               cram-bullet-reasoning-utilities
```
gaya-: I don't like this dependency. I prefer using the utils only for hacking and high-level application code.
But for the libraries I think we should keep the dependencies to the minimum and just use direct btr stuff.
For moving objects, please use `(setf (pose (btr:object btr:*current-bullet-world* :my-obj-name)) my-new-pose)`. For deleting objects I don't remember the exact `btr` function, but you can either use `(btr:retract btr:object ?world ...` Prolog predicate or see how the `retract` is implemented in `btr` and see which low-level function that predicate is calling.

amar-fayaz: Hi, I'm working towards getting this dependency removed.
I've gotten the move and delete working but my spawn-object (used in tests) is not working. 
Just to see its behavior, I tried copy-pasting the spawn-object definition from btr-utils and that hasn't helped either. Do you know any alternatives where this is used?

gaya-: Hmm, what did you try?
Why not use the stuff from the Bullet tutorial then:
```lisp
CL-USER> (prolog:prolog '(and (btr:bullet-world ?world)
                              (assert (btr:object ?world :mesh mug-1 ((0 0 2) (0 0 0 1))
                                                  :mass 0.2 :color (1 0 0) :mesh :mug))))
```

http://cram-system.org/tutorials/intermediate/bullet_world#spawn_and_change_objects_in_the_world

You have to provide a mass but not the color.
Let me know if that doesn't work.

amar-fayaz: I removed the dependency in the latest change. 1 test case now breaks due to some issue, which I haven't figured out yet. The object in that particular test alone acts as if there is no floor and keeps falling endlessly. I have commented that test out alone and investigating it meanwhile.

artnie:
tl;dr
Recommended changes:
* change the error at the end of the correction function to a warning.
* do the simulation, re-spawning, etc in a copy of the `*current-bullet-world*` by using `(bullet:copy-world btr:*current-bullet-world*)`. In this copy we are able to try if the correction is feasible in the first place, then apply the new pose to the detected object in the  `*current-bullet-world*`.

Optional:
* calculate where the object is moving by subtracting the x,y,z values, to get a new vector. Move the object in the direction opposing this calculated vector. 

A detected object is first spawned normally in the bullet world, then you simulate for a while and measure the distance between the original pose and the one where the object is after simulation. If this distance is big enough, re-spawn the object just a wee apart from its originally perceived pose, then simulate again. The correction moves the object into the direction opposing to the one is takes during simulation, so if the object is moving down during simulation it will be re-spawned slightly above its original position.

I see a bit of an issues here. This re-spawning of a wildly moving object with the little offset is done to all the axis, not only towards positive Z, depending on the direction the object moved during simulation. This might lead to unwanted/unnecessary correction of the object. On the other hand it will prevent a newly detected object to fall from an edge, because the object would be re-spawned towards the center of its supporting surface. This kind of correction only fires, if the original pose and the one after simulation deviate by a distance of at least 10cm, which is nice.

A standard Lisp Error is raised, if the correction isn't able to stabilize the detected object, meaning, when the object moves too far after being re-spawned and simulated. This should not happen, since a standard error will interrupt the program flow. Instead, further correction should be applied, or the object must be rejected eventually.

Such further correction could include checking the surface the object is in contact with when it is spawned. Also the direction of the correction could be done by calculating the vector between the original position and the one after a short simulation.

gaya-:
Got replaced with https://github.com/cram2/cram/pull/139

PR #131, opened 2019-08-21
Holes in global costmap after mapping a costmap rotated on it 2

tlipps:
In this PR the holes in the global costmap are fixed. The holes originated from the rotated input costmaps (learned from the cram_learning_framwork) which were mapped with the standard rotation matrix and landed sometimes in adjacent fields and created so holes in the global costmap. 
The problem was solved by mapping not only the point alone but the middle of the point too. This solution seemed to work pretty well and can be tested for the riskiest angles with the following code snippet: 
```
(loop for i from 2 to 4 do
          (let ((cm
                    (make-instance
                        'location-costmap:location-costmap
                      :resolution 0.02
                      :origin-x -6
                      :origin-y -6
                      :height 12.0
                      :width 12.0)))
              (costmap:register-cost-function 
               cm
               (costmap:make-matrix-cost-function 
                0.0 0.0 1.0
                (make-array '(4 4)
                            :element-type 'double-float
                            :initial-element 0.01d0)
                (/ pi i))
               :hello)
            (sleep 0.2)
            (costmap:costmap-samples cm)))
``` 
Originated from https://github.com/cram2/cram/pull/126. Was tested again with above code snippet.

### DIFF
--- a/cram_common/cram_location_costmap/src/cost-functions.lisp
+++ b/cram_common/cram_location_costmap/src/cost-functions.lisp
@@ -175,15 +175,28 @@ in a value of 1.0"
             (occupancy-grid-put-mask col row result-grid padding-mask :coords-raw-p t))))
       (make-occupancy-grid-cost-function result-grid :invert invert))))
 
-(defun rotate-point-around-another-point (x y sin-theta cos-theta
+(defun rotate-point (x y sin-theta cos-theta &optional rotation-point-x rotation-point-y)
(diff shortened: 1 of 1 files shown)


gaya-:
Cleaned up in https://github.com/cram2/cram/commit/2c9873ecc31b291da3f4f60c0957bf97e6b2d2c1
Nice trick with these middle points.

PR #134, opened 2019-09-10
Fall School Tutorial Blank File For Work

amar-fayaz:
Adding a blank file to the project so that the participants can add and test their methods without completely relying on the REPL.

There was an existing file which wasn't a part of the package. Removed the code inside it and added it to the package.


gaya-:
Cherry-picked, thanks!
Please do git reset --hard origin/master or so, your branch diverged from the master branch. But you might end up losing data, so backup first.

PR #136, opened 2019-09-18
[env-man] Fix clipping and making relative of the opening/closing angle.

cpollok:
When manipulating the environment the joint state values get shifted to other values (still the same angle but shifted by 2pi). This lead to bugs in calculating the angle for the trajectory generator.
So now the state values used in the get-relative-distance and clip-distance functions are put through the modulo operator to get the positive angle value between 0 and 2pi.


gaya-:
Awesome, glad this is fixed now. So what we learned from this is that, in addition to unit tests, we also always need to have at least very few integration tests, to see how the new feature fits into the rest of the code.

PR #112, opened 2019-05-07
Attachment: collision flags, detach-all-objects and unidirectional/loose attachments for items, object won't collide with attachments in robot arm

tlipps:
## New Stuff
Added and refactored like mentioned in the [trello card](https://trello.com/c/szsA84QM/54-attaching-objects-to-objects-have-general-functions-that-attach-objects-to-objects-and-objects-to-robot-links):

- abstract attach-object/detach-object functions for robots and items with proper collision flags

- functions for attaching items loose or one item to more items directly

- functions for handling collision flags, since things made of items attached to each other collide

- detach-all-objects for items

 Unidirectional/loose attachments were implemented [here](https://github.com/cram2/cram/pull/112/commits/bd045449a07710b6f10175dd7659ed1cbd164f11).

It is possible that items can be attached with more than one item. So this should be clear for unidirectional/loose attachments too. This was implemented [here](https://github.com/cram2/cram/pull/112/commits/152c5d53b89d878719beeee40696b9a44fa78db7) and in the above [commit](https://github.com/cram2/cram/pull/112/commits/bd045449a07710b6f10175dd7659ed1cbd164f11).

Since moving the robot arm with an object with attachments could lead to collisions (in low-level.lisp in package cram_urdf_projection) [a method](https://github.com/cram2/cram/pull/112/commits/a8d06feaf7f13e662109e487e89015c8f5cdf2eb) was added in robot-model-utilities.lisp, which checks, if the moving object has attachments and if so doesn't throw collision errors. To test this one can just execute the demo function in cram_boxy_assembly_demo or execute the unit tests below.

After all in this PR are contributions made to the cram_boxy_assembly_demo package too.

## Tests
All of the above can be tested with lisp-unit. For that execute following lines:
```
(ros-load:load-system "cram_bullet_reasoning" :cram-bullet-reasoning-tests)
(in-package :btr-tests)
(roslisp-utilities:startup-ros)
(lisp-unit:run-tests :all :btr-tests)
```
In asd file the cram_pr2_description is loaded.

PR #137, opened 2019-09-24
Formated test code in btr-tests and more documentation in btr

tlipps:
This PR is a small extension to the PR https://github.com/cram2/cram/pull/112. In here the missing small fixes of above PR were added:

* format the code in the package btr-tests in cram_bullet_reasoning (mentioned: https://github.com/cram2/cram/pull/112#discussion_r316258841)
* more understandable documentation for functions in items.lisp in cram_bullet_reasoning (mentioned: https://github.com/cram2/cram/pull/112#discussion_r316365836)


gaya-:
Thomas's comment explaining `remove-loose-attachment-for`:

"The idea of btr::remove-loose-attachment-for is that it removes all the (indirect) loose attachments from an object. So for this it gets first all object names with loose attachments for the given object. When some loose attached objects were found they will be detached from the given object. So we have e.g. a mug which was attached loose on the tray. If btr::remove-loose-attachment-for gets the mug object as input it should now remove the loose attachment from the mug to the tray.

Now comes the recursive part. If the given object has attached-objects and there is something in it (length > 0), we push it to already-visited. The idea is now if we have a thing made of attached items (e.g. the toy plane) we traverse trough each part of the plane (via the attachments) to check if any part was connected loose with e.g. a tray. This is because if we attach the plane via the e.g. upper-wing of the toy plane to something, we want that the loose attachment of the e.g. rear-wing to the tray gets removed too. Otherwise, if the tray moves the plane moves with it. And to remove the loose attachment recursive the following part of btr::remove-loose-attachment-for is needed.

So btr::remove-loose-attachment-for was called with e.g. the upper wing of the toy plane. Since it has attached-objects btr::remove-loose-attachment-for gets called with every attached object. Since the attached objects of upper-wing contain upper-body and the attached objects of upper-body contain upper-wing, (member (name attached-object) already-visited) checks if upper-wing or upper-body was already visited with remove-loose-attachment-for. This is done to prevent a continuous loop.

At the end the method checks if it was called with the object with which btr::remove-loose-attachment-for was called first. So in our example upper-wing. If that's the case already-visited will be set to an empty list since all (indirect) attached objects of upper-wing were traversed before.

I think we discussed having these closures is problematic since with a new compilation of the lisp code these are initialized again. So I thought about adding a defvar *updated-attachments* instead, because defvar initialized variables are initialized only once. I wanted to discuss this with you too since in some methods I used now closures and in some others I used defvar."

gaya-:
As this is a PR that cleans up #112, I will wait for @tlipps to get rid of the `defvar` and replace it with a closure if possible.

tlipps:
> As this is a PR that cleans up #112, I will wait for @tlipps to get rid of the `defvar` and replace it with a closure if possible.

Done. I reran the tests from #112 and everything passed. I had to add a getter for the closure to test the function using the defvar.

gaya-:
Merged in 218bdf9

Issue #82, opened 2018-12-17
Test pick and place plans in supermarket environment


gaya-:
For that

1. Make a URDF of a small supermarket shelf (or ask Simon if he has one already)
2. Spawn the URDF and a robot of choice into Bullet world
3. Define grasping and placing poses for items in the supermarket
4. Spawn a mesh of a shopping basket
5. Define placing poses on the basket
6. Test picking object from supermarket shelf and placing it into the basket.
7. Define placing poses on supermarket shelf
8. Test picking objects from shopping basket and placing them onto shelf.

gaya-:
Implemented in cram_pr2_shopping_demo on boxy and boxy-melodic branches.

PR #140, opened 2019-11-07
Shopping demo

Tigul:
Implemented fetch and deliver plans into the sjopping demo


cram_pr2/cram_pr2_shopping_demo/launch/world.launch
```diff
@@ -4,6 +4,10 @@
     <!-- uploading the shelf to the param server -->
     <param name="shelf_description" command="$(find xacro)/xacro.py
         '$(find cram_pr2_shopping_demo)/resource/shelf.urdf'"/>
+    <!-- uploading basket urdf to param server -->
+    <param name="basket_description" command="$(find xacro)/xacro.py
+        '$(find cram_pr2_shopping_demo)/resource/basket_urdf.urdf'"/>
+
```
gaya-: If we are using primitive shapes now, we don't need a basket urdf, I would get rid of this. If you want to keep it, at least comment it out and write a comment saying that the URDF is not used anymore for X and Y reasons.

cram_pr2/cram_pr2_shopping_demo/resource/basket_urdf.urdf
```diff
@@ -0,0 +1,149 @@
...
+        </visual>
+    </link>
+
+    <joint name="handle_middle_to_bottom_joint" type="fixed">
+        <origin xyz="0 0 0.27"/>
+        <parent link="bottom_link"/>
+        <child link="handle_middle_link"/>
+    </joint>
+
+</robot>
```
gaya-: Why were the stl files removed? Like `denkmit.stl`, `dove.stl` and `heitmann.stl`?

cram_3d_world/cram_bullet_reasoning/src/items.lisp
```diff
@@ -171,6 +171,85 @@ The name in the list is a keyword that is created by lispifying the filename."
                                                                       mass)
   (add-object world :mesh name pose :mass mass :mesh :mug))
 
+(defun make-rectangle-shape (lenght width height)
```
gaya-: `make-rectangle-shape` is a weird name for this function. It's not really a rectangle that you are creating.

I think, it would probably be best to just put all these parts of the basket into one function. So combine `make-rectangle-shape`, `make-handle-shape` and `make-basket-shape` together and put it into `add-object :basket`.
I don't see us using any of these functions for any other object at this moment. For now, it is only confusing what this `make-rectangle-shape` is used for. But when you know that this is a part of a basket, it is much clearer.

cram_pr2/cram_pr2_shopping_demo/resource/basket_urdf.urdf
Tigul: They weren't removed, I decimated the shaped and because of this the files changed

gaya-:
Please add the highres meshes back to resource.

gaya-:
Solves #82

Tigul:
# Current State
Putting Objects in the basket works and they don't fall to the ground anymore.
However, because the location designator contains an "on" key the btr-spatial-relation costmap is spawned. Because of this the prolog statements for this costmap are commented out at the moment.
# Dependencies
This PR depends on @tlipps branchens "drop-it-to-attach" and "new-boxy-new".

gaya-:
Merged manually in 31fe29c.
Looks good, especially after the cleanup.
Thanks a lot, it's a fun demo.

PR #141, opened 2019-11-07
Compound meshes for items

artnie:
This PR is mostly based on Mihais #77 PR, implementing compound shapes for items other than robots and providing a central function for any kind of mesh to be loaded as compound.
My changes are:
* removed the visual component
* straighten the loading and representation of mesh files by introducing a default parameter, spawning a mesh as compound any time it is possible and keeping it a single convex shape if the mesh only has one component
* added the test to the other tests for `btr` and adjusted it for `lisp-unit`


mpomarlan:
Yay!

PR #145, opened 2019-11-27
Quadtree Library

tlipps:
In this PR is a new library called quadtree included which implements Quadtrees.  
This library brings following methods:

- Creating of Quadtrees represented as struct with object, subtrees, boundary and depth parameter
- Querying of Quadtrees with a possibility to return the quried path
- Inserting of objects with excluding middlepoints since these increment the depth of a Quadtree
- Set and get object values (represented as the parameter z in a 3d-vector) in Quadtrees
- Map functions over Quadtrees
- Import a matrix into an empty Quadtree
- Basic minimizing: meaning if the leaves have all the same value or are all empty these leaves can be removed and their parent gets updated
- Rotation: boundaries of quadtree will be rotated

This library is not unit-tested and may therefore contain bugs. Moreover, the rotation implementation should be changed: the rotation of boundaries is way to heavy since these have to be properly aligned when merging Quadtrees. Therefore, these should be static.

### DIFF
--- a/cram_3rdparty/quadtree/CMakeLists.txt
+++ b/cram_3rdparty/quadtree/CMakeLists.txt
@@ -1,0 +1,202 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(quadtree)
+
+## Compile as C++11, supported in ROS Kinetic and newer
+# add_compile_options(-std=c++11)
+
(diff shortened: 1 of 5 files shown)


gaya-:
Cheery picked, thanks!

PR #147, opened 2019-12-05
Automatically infer object likely locations and destinations

tlipps:
**New Stuff**:

- Implemented functions `get-object-likely-destination/location` in `cram_manipulation_interfaces:manipulation-interfaces.lisp`
- Created `defmethods :heuristic 20` in `cram_object_knowledge:household.lisp` 
- Integrated `get-object-likely-location/destination` in `cram_fetch_deliver_plans`

**Testing**:

Since mostly location designators were moved from `cram_pr2_pick_place_demo:demo.lisp` to `cram_object_knowledge:household.lisp` through `get-object-likely-location/destination,` I reran the `demo-random` in `cram_pr2_pick_place_demo:demo.lisp.`

terminal:
`roslaunch cram_pr2_pick_place_demo sandbox.launch`

roslisp_repl:
```
(ros-load:load-system "cram_pr2_pick_place_demo" :cram-pr2-pick-place-demo)
(ros-load:load-system "cram_pr2_description" :cram-pr2-description)
(in-package :demo)
(roslisp-utilities:startup-ros)
(urdf-proj:with-simulated-robot (demo-random nil))
```

### DIFF
--- a/cram_common/cram_object_knowledge/src/household.lisp
+++ b/cram_common/cram_object_knowledge/src/household.lisp
@@ -325,3 +325,440 @@
   :2nd-pregrasp-offsets `(0.0 0.0 ,*bowl-pregrasp-z-offset*)
   :lift-offsets `(0.0 0.0 ,*bowl-pregrasp-z-offset*)
   :2nd-lift-offsets `(0.0 0.0 ,*bowl-pregrasp-z-offset*))
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;; location ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defmethod man-int:get-object-likely-location :heuristics 20
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-setting))
+     (?object-type (eql :bowl)))
+
+  (desig:a location
+           (on (desig:an object
+                         (type counter-top)
+                         (urdf-name sink-area-surface)
+                         (owl-name "kitchen_sink_block_counter_top")
+                         (part-of ?kitchen-name)))
+           (side left)
+           (side front)
+           (range-invert 0.5)))
+
+(defmethod man-int:get-object-likely-location :heuristics 20
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-setting))
+     (?object-type (eql :spoon)))
+  
+  (desig:a location
+           (in (desig:an object
+                         (type drawer)
+                         (urdf-name sink-area-left-upper-drawer-main)
+                         (owl-name "drawer_sinkblock_upper_open")
+                         (part-of ?kitchen-name)))
+           (side front)))
+
+(defmethod man-int:get-object-likely-location :heuristics 20
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-setting))
+     (?object-type (eql :knife)))
+  
+  (desig:a location
+           (in (desig:an object
+                         (type drawer)
+                         (urdf-name sink-area-left-upper-drawer-main)
+                         (owl-name "drawer_sinkblock_upper_open")
+                         (part-of ?kitchen-name)))
+           (side front)))
+
+(defmethod man-int:get-object-likely-location :heuristics 20
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-setting))
+     (?object-type (eql :plate)))
+
+  (desig:a location
+           (on (desig:an object
+                         (type counter-top)
+                         (urdf-name sink-area-surface)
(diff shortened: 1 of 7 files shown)


cram_common/cram_object_knowledge/src/household.lisp
```diff
@@ -325,3 +325,440 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;; knife ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defmethod man-int:get-object-likely-destination :heuristics 20 
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-setting))
+     (?object-type (eql :knife)))
+  
+  (desig:a location
```
gaya-, line 552: knife should rather be right of plate

cram_common/cram_object_knowledge/src/household.lisp
```diff
@@ -325,3 +325,440 @@
...
+             ;; (side back)
+             )))
+
+(defmethod man-int:get-object-likely-destination :heuristics 20
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-cleaning))
+     (?object-type (eql :breakfast-cereal)))
+
+  (desig:a location
```
gaya-, line 699: it goes to the drawer, according to the milestone scenario

cram_common/cram_object_knowledge/src/household.lisp
```diff
@@ -325,3 +325,440 @@
...
+                         (owl-name "kitchen_sink_block_counter_top")
+                         (part-of ?kitchen-name)))))
+
+(defmethod man-int:get-object-likely-location :heuristics 20
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-setting))
+     (?object-type (eql :breakfast-cereal)))
+
+  (desig:a location
```
gaya-, line 426: should switch to left vertical drawer according to milestone scenario. however, test beforehand if opening the drawer is possible. should be.

cram_common/cram_object_knowledge/src/household.lisp
```diff
@@ -325,3 +325,440 @@
...
+             (far-from (desig:an object (type ?other-object)))
+             (for (desig:an object (type ?object-type))))))
+
+(defmethod man-int:get-object-likely-destination :heuristics 20
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-cleaning))
+     (?object-type (eql :milk)))
+
+  (desig:a location
```
gaya-, line 727: this should be the fridge

gaya-: or the trash can

gaya-: this depends on the weight of the object or its "status" like empty or not empty. Thus, the query would need additional input parameters. So we 
  * either &allow-other-keys,
  * or pass the whole object designator
  * or read the object state from the world state, which is not quite right, as we don't know which specifically milk box are we talking about, what if there is an empty milk box and a non-empty milk box in the world. thus, we have to identify the object somehow in the query interface. in which case, if we at least pass its name, we might as well pass another argument, such as status or state or weight or fill-level or ...

cram_common/cram_object_knowledge/src/household.lisp
```diff
@@ -325,3 +325,440 @@
...
+                         (part-of ?kitchen-name)))
+           (for (desig:an object (type ?object-type)))))
+
+(defmethod man-int:get-object-likely-destination :heuristics 20
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-cleaning))
+     (?object-type (eql :bottle)))
+
+  (desig:a location
```
gaya-, line 757: Throwing bottles into the sink is kind of weird. So I'd say either trash or back onto the sink area surface.

cram_common/cram_object_knowledge/src/household.lisp
gaya-: made a trello card

cram_common/cram_object_knowledge/src/household.lisp
```diff
@@ -325,3 +325,440 @@
...
+(defmethod man-int:get-object-likely-destination :heuristics 20
+    ((?kitchen-name (eql :kitchen))
+     ?human-name 
+     (?context (eql :table-cleaning))
+     (?object-type (eql :bowl)))
+
+  (desig:a location
+           (on (desig:an object
+                         (type area-sink)
+                         (urdf-name sink-area-sink)
```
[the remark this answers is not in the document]
gaya-: urdf-name is correct, owl name might not even exist... but we don't perceive objects there, we only throw them in without perceiving, so it's ok

gaya-:
Merged manually.

PR #150, opened 2019-12-12
Drop it to attach

tlipps:
## Functionality & Tests

This PR adds the generic function `get-z-offset-for-placing-distance` making it possible to define z offsets for placing tasks. As an example one can add the following line
```

(format t "dropping from ~A to safe pose ~A" (btr:pose btr-object) object-in-map-pose)
(break)

```
 in `cram_manipulation_interfaces:event-handlers.lisp` line 130 and execute the `demo` in `cram_boxy_assembly_demo`. The object `bottom-wing' will be placed above the `chassis` object and after that will be "simulated" in its real position. The difference can be seen in the z values of both poses.

One more example is shown in https://github.com/cram2/cram/pull/140. 

## Bug: lost collision-information 

https://github.com/cram2/cram/pull/140 revealed a bug too. If a non-stable object was grasped from the robot and placed in the basket, it would fall although it is attached to the basket. The reason for this was that the "non-stable" collision-information was reset in the (detach-object rob obj) call. Because the attachment from the object to the basket was made before the detachment, the object was not set again to be stable in btr.

### Fix
 
To keep the collision-information from the object valid and consistent, code was written in this PR (https://github.com/cram2/cram/pull/150/commits/667b90b214d279c03596c606b0b5bfe23c61ca33). Since the bug affects only items, these changes were made in items.lisp and objects.lisp.

#### items.lisp:

* attach-object: Since items are not connected through their attached-objects to robots, items try to get their collision-information from the robot if they attach to another item.
* detach-object: Detaching other-object from object before resetting the collision-information. 

#### object.lisp

* create-static-collision-information: If the item i has no attached objects, the collision-information from btr gets gathered. Else the attached-objects of i will be checked. If i was marked as static in any of i's attached objects, it was once a static object.

* reset-static-collision-information: If the item got attached-objects, it should be static, else the collision-information gets loaded in the object.

### DIFF
--- a/cram_3d_world/cram_bullet_reasoning/src/items.lisp
+++ b/cram_3d_world/cram_bullet_reasoning/src/items.lisp
@@ -354,8 +354,17 @@ and the other one is loosly attached. "
       (if (equal (car (last already-visited)) (name object))
           (setf already-visited '())))))
 
+(defun object-attached-to-robot-p (object)
+  (when object
+    (assoc (name object) (btr:attached-objects (get-robot-object)))))
+
+(defun get-collision-information-from-robot (object)
+  (declare (type item object))
+  (list (caddr (assoc (name object) (attached-objects (get-robot-object))))))
+
 (defmethod attach-object ((other-object item) (object item)
-                          &key attachment-type loose skip-removing-loose link grasp)
+                          &key attachment-type loose
+                          skip-removing-loose link grasp)
   "Attaches `object' to `other-object': adds an attachment to the
 attached-objects lists of each other. `attachment-type' is a keyword
 that specifies the type of attachment. `loose' specifies if the attachment
@@ -376,7 +385,11 @@ attaching more objects unidirectional and should be for this T. See
               (cons
                (list (make-attachment :object (name object)
                                       :attachment attachment-type))
-               (create-static-collision-information object)))
+               ;; Since robot objects are not in the attached-objects
+               ;; list of items, this has to be copied manuelly:
+               (if (object-attached-to-robot-p object)
+                   (get-collision-information-from-robot object)
+                   (create-static-collision-information object))))
         (slot-value other-object 'attached-objects))
   (push (cons (name other-object)
               (cons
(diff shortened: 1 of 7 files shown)


gaya-:
This PR might not work for table setting, where we check for goal pose stability before placing.

cram_3d_world/cram_bullet_reasoning/src/items.lisp
```diff
@@ -354,8 +354,17 @@ and the other one is loosly attached. "
       (if (equal (car (last already-visited)) (name object))
           (setf already-visited '())))))
 
+(defun object-attached-to-robot-p (object)
```
gaya-, line 357: This already exists, it's called OBJECT-ATTACHED

cram_3d_world/cram_bullet_reasoning/src/items.lisp
```diff
@@ -354,8 +354,17 @@ and the other one is loosly attached. "
...
+  (when object
+    (assoc (name object) (btr:attached-objects (get-robot-object)))))
+
+(defun get-collision-information-from-robot (object)
```
gaya-, line 361: This is GET-COLLISION-INFO from the FLET of DETACH-OBJECT from robot

cram_common/cram_manipulation_interfaces/src/manipulation-interfaces.lisp
```diff
@@ -81,3 +81,11 @@ is considered closed.
 `container-designator' is a designator describing the container.")
   (:method :heuristics 20 (container-name)
     nil))
+
+(defgeneric get-z-offset-for-placing-distance (other-object object attachment)
```
gaya-, line 85: This should not be a general knowledge query. It's only used in our heuristics calculation for the placing trajectory. So it's rather heuristics-engine specific. So move this to where the get-object-type-grasp-transforms is defined, i.e. trajectories.lisp i guess.

cram_common/cram_manipulation_interfaces/src/package.lisp
```diff
@@ -55,6 +55,7 @@
    #:get-location-poses
    #:get-container-opening-distance
    #:get-container-closing-distance
+   #:get-z-offset-for-placing-distance
```
gaya-, line 58: under ;; trajectories

cram_common/cram_manipulation_interfaces/src/manipulation-interfaces.lisp
gaya-: Rename and exchange parameter order.

cram_common/cram_manipulation_interfaces/src/package.lisp
gaya-: rename

cram_common/cram_manipulation_interfaces/src/trajectories.lisp
```diff
@@ -304,9 +304,31 @@ Gripper is defined by a convention where Z is pointing towards the object.")
...
+         (z-offset (get-z-offset-for-placing-distance
+                    maybe-other-object-type
+                    object-type
+                    maybe-attachment))
          (oTg-std
-           (get-object-type-to-gripper-transform
-            object-type object-name arm grasp)))
+           (let ((tmp-oTg-std (get-object-type-to-gripper-transform
+                               object-type object-name arm grasp)))
+             (cl-tf:transform->stamped-transform
```
gaya-, line 320: copy-transform-stamped or smt

cram_3d_world/cram_bullet_reasoning/src/objects.lisp
```diff
@@ -358,16 +358,56 @@
       (loop for body in (rigid-bodies object)
             collecting (make-collision-information
                         :rigid-body-name (name body)
-                        :flags (collision-flags body))
-            do (setf (collision-flags body) :cf-static-object))))
+                        :flags (or (when (not (attached-objects object))
+                                     (collision-flags body))
+                                   (if (object-static-in-past-p object)
+                                       '(:cf-static-object)
+                                       NIL)))
```
gaya-, line 365: It should not be NIL, it should be the old collision information that is stored. The collision information should simply be taken over if the status of "attached" hasn't changed for the object. So, if it is already attached, perhaps nothing should happen at all. And the same during resetting, if it is still attached, nothing should happen (if that doesn't work, the stored collision information should be taken over).

gaya-:
@tlipps There were two very distinct things happening in this PR: the get-z-offset method and the bugfix for collision information. The bugfix is somewhat related to the dropping but not really. In the future, please create 2 different PRs if the features are so separate from each other.

gaya-:
Merged manually, thanks. I did not like the solution with the collision flags, though, so there is a new card on Trello about it now.

PR #153, opened 2020-01-16
Fix env man distance calculation

cpollok:
Problem was that the distance calculation did not distinguish between prismatic and revolute joints. So a mod 2*pi was performed on the joint state of prismatic joints. (this was intended for revolute joints, because bullet assigns them negative numbers sometimes, to keep them in the 0-2pi range.)
This lead to very large distances for the opening and closing of prismatic joint containers.

~~This also fixes a little bug in the env man plan, which was not closing the gripper, because the wrong designator was used.~~

This also makes the object designator needed in the gripping action optional, so that the env-man plan can still use gripping and have the benefits without needing to provide an object that is being gripped. (Because attaching a container to the gripper would be not good.)

Also fixes a small typo in the environment manipulation plan.

### DIFF
--- a/cram_3d_world/cram_urdf_environment_manipulation/src/environment.lisp
+++ b/cram_3d_world/cram_urdf_environment_manipulation/src/environment.lisp
@@ -238,6 +238,25 @@ Using a default (1 0 0)."
         (cl-urdf:name joint))
        (* 2 pi)))
 
+(defun get-connecting-joint-state-secure (container-name btr-environment)
+  (let* ((joint (get-connecting-joint
+                 (get-container-link container-name
+                                     btr-environment)))
+         (type (find-container-joint-type-under-joint joint))
+         (state (btr:joint-state
+                     (btr:object btr:*current-bullet-world*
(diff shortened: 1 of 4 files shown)


cram_3d_world/cram_urdf_environment_manipulation/src/environment.lisp
```diff
@@ -238,14 +238,33 @@ Using a default (1 0 0)."
...
+(defun get-connecting-joint-state-secure (container-name btr-environment)
+  (let* ((joint (get-connecting-joint
+                 (get-container-link container-name
+                                     btr-environment)))
+         (type (find-container-joint-type-under-joint joint))
+         (state (btr:joint-state
+                     (btr:object btr:*current-bullet-world*
+                                 btr-environment)
+                     (cl-urdf:name joint))))
+    (if (eq type :rotational)
```
gaya-, line 250: It's not rotational, it's revolute, how does this even work.

gaya-:
Merged manually.

PR #154, opened 2020-01-16
[env-man] Use range costmap instead of gaussian for reachability.

cpollok:
This changes the costmaps for both prismatic and revolute joint containers to use the range costmap instead of a gaussian for reachability. This makes much more sense and should make the costmaps more stable. But it might be necessary to make sure the costmap metaparameters are sound (minimal and maximum reaching distance).
Testing included opening drawers and the fridge with both arms of PR2 and Boxy. Especially the left upper drawer with the right arm and the trash drawer with the left arm were tested thoroughly.

This PR also increases the PR2's in-reach-distance by 10cm to make the costmap work.

### DIFF
--- a/cram_3d_world/cram_urdf_environment_manipulation/src/costmaps.lisp
+++ b/cram_3d_world/cram_urdf_environment_manipulation/src/costmaps.lisp
@@ -118,13 +118,13 @@ environment, in which it can be found, respectively."
           (line-equation-in-xy neutral-point manipulated-point)
         (let* ((P (cl-transforms:make-3d-vector x y 0))
                (dist (line-p-dist a b c P))
-               ;;(dist-p (line-p-dist-point a b c P))
+               (dist-p (line-p-dist-point a b c P))
                )
           (if (and
                (< dist (+ (/ width 2) padding))
                ;; Commenting this out for now, so there won't be poses in front of the drawer.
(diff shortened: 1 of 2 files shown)


gaya-:
I load the pr2 pick place demo, load the pr2 description, start the launch file, start the ros node and execute:

```lisp
(urdf-proj:with-projected-robot
  (perform (an action
               (type accessing)
               (arm left)
               (location 
                (desig:a location
                         (in (an object
                                 (type drawer)
                                 (urdf-name sink-area-left-upper-drawer-main)
                                 (part-of kitchen))))))))
```
Get a tiny costmap, which doesn't even result in one sample.
This PR doesn't work, sorry.

cpollok:
This now also increases the PR2's in-reach-distance to 1.0 so the costmap works.
I developed this on 18.04 but the PR was originally on 16.04, but I think leaving it to be pulled to the boxy branch is easier than switching.

gaya-:
@cpollok I don't get it though why it worked with the left hand so well but not the right hand. Why the costmap is not symmetrical...

gaya-:
Merged manually, thanks.

gaya-:
@cpollok Ah, they are symmetric, it's just that our kitchen is not. Get it, all good.

PR #158, opened 2020-02-17
Integrated world-state-detecting in perceiving

Tigul:
# Added with this PR
* Integrated world-state-detecting in the perceiving plan 
* Added a `counter` key to the perceiving designator 
* Implemented a counter variable in `fetch` which counts the failed perceiving attempts
* Changed world-state-detecting, now only the type of the object is enough.

### Integration world-state-detecting with fetch and deliver
In `fetch` a counter was implemented which counts how often a `perception-object-not-found` error was thrown.
In the Prolog statements which resolve the perceiving designator it will be checked if the given object designator contains a `name` key. If this is the case occluding objects will be determined. 
In the `perceive` plan it will firstly be tried to detect the object using `detecting`, if this throws an perception-object-not-found error the error will be catched and when there are occluding objects and the counter reached 0 the world-state-detecting designator will be used.

### World-state-detecting with only a type
using the `btr:get-objects-for-type` function all objects for the given type will be queried, then all occluding objects will be determined for every object of the given type. If there are it means the object is in the field of view (FOV) of the robot and occluded by other objects. This returns a list of all objects which are occluded but in the FOV of the robot. 
If the resulting list, of occluding objects, is empty a `perception-object-not-found` error will be thrown


cram_common/cram_mobile_pick_place_plans/src/atomic-action-plans.lisp
```diff
@@ -348,26 +348,50 @@ equate resulting designator to the original one."
...
                       resulting-designators)
               (progn
                 (cram-occasions-events:on-event
                  (make-instance 'cram-plan-occasions-events:object-perceived-event
-                   :object-designator resulting-designators
-                   :perception-source :whatever))
+                                :object-designator resulting-designators
+                                :perception-source :whatever))
                 (desig:equate ?object-designator resulting-designator)))
           resulting-designator)))))
```
artnie: Remove extra parantheses to this: `resulting-designator)`

sunava:
integrated here: [PR259](https://github.com/cram2/cram/pull/259) Thank you!

PR #159, opened 2020-02-19
[env-man] Fix doc strings in math, environment and trajectories lisp files

cpollok:
Adds or improves the doc strings in the mentioned files. This was done mainly to improve understandability in the trajectories.lisp file. So now all transform parameters are described as to how they should look and what they contain.


gaya-:
@cpollok This looks good but please use type declarations for input arguments.
```lisp
(declare (type cl-urdf:joint my-joint-argument my-other-joint-argument)
              (type number some-joint-state))
```

cpollok:
I used declarations and some the clauses for the return values.
I am not sure about the commits in this though. Can/Should I just remove the two revert commits and/or the Merge commit of boxy-melodic into boxy? All of those are only there because this started developement on the boxy branch, but I now am on boxy-melodic.
@gaya- If you can tell me what I should do about those, I can try to do it myself.

gaya-:
I'll just cherry-pick I guess. Don't worry.

gaya-:
Merged manually, thanks.

PR #160, opened 2020-02-19
Hsrb

Vanessa-rin:
I implemented the hsr into the cram framework.
I've called hsrb since there are some differences between hsr-c and -b. 
Since now he is only able to grasp the cylinder for everything else we need to add more specific knowledge.


gaya-:
Merged manually in 1c557d6. Thanks!

PR #163, opened 2020-03-19
[external_interface]Adding mechanism to collision check during ik resampling (for torso, etc.)

amar-fayaz:
Adding a new ik service caller that does resampling as well as collision check. 
The new method works similar to the resampling version but now takes 2 more arguments
1. The collision check function
2. The collision check arguments as a list.

The appropriate collision check function can be passed down the chain and can be used to look for new solutions of ik during resampling. Any collision check error is only used to force the ik solver to sample the next solution and is not bubbled up.

Added the corresponding dependencies and exposed the API in the package, and using the said API in the urdf_projection package.

### DIFF
--- a/cram_3d_world/cram_urdf_projection/src/low-level.lisp
+++ b/cram_3d_world/cram_urdf_projection/src/low-level.lisp
@@ -807,7 +807,7 @@ with the object, calculates similar angle around Y axis and applies the rotation
                                   ?torso-link ?left-ee-frame ?left-arm-joints
                                   ?torso-joint ?lower-limit ?upper-limit
                                   :collision-check-function #'perform-collision-check
-                                  :collision-check-arguments (list collision-mode left-tcp-pose right-tcp-pose)))
+                                  :collision-check-arguments `(,collision-mode ,left-tcp-pose ,right-tcp-pose)))
       (multiple-value-bind (right-ik right-torso-angle)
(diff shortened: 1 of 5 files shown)


amar-fayaz:
Updated with the changes as per the call

gaya-:
@amar-fayaz There's an infinite recursion somewhere. When the IK solver gives a colliding solution, somewhere there...

gaya-:
@amar-fayaz ok, so, first, you obviously didn't even test this, because with-failure-handling propagates the failure, if it isn't handled, which in your implementation is the case. And, second, you told me that one has to keep in mind to set the calculated joint state before collision check and reset the world afterwards, and it's not in the PR. So, basically, the collision check always just checks the same joint state, which makes no sense...

gaya-:
@amar-fayaz or you tested it in a situation where there could not be any collision, which is the only use case where the infinite recursion doesn't happen.

gaya-:
Fixed in 6527014.
@amar-fayaz I guess you didn't test it because you thought it is only used for Boxy?
It is, in fact, used for both boxy and pr2, so one can test the same with pr2.
I did get, in fact, a lot of unhandled memory faults with boxy on my 18.04 set up.
Which is good news, because it means, now I can reproduce your boxy problem.
And bad news, because these memory faults happen way too often...

PR #167, opened 2020-04-24
IK solving with chains including fixed joints

amar-fayaz:
Added a README file with example to call the method with.
Added a workaround on the python script to use zeros as seed state if no valid solution is arrived with the initial seed values.


gaya-:
Very nice. The fall back with zeros helps even for the 7-joint chain. So I guess Boxy will become more flexible and will reach better now. Thanks for the PR!

PR #171, opened 2020-05-12
Resampling code for multiple axes

amar-fayaz:
[Trello](https://trello.com/c/JVMVA1Yq)

Allowing Base Resampling along with torso resampling of the robot.  
Created a new macro which allows nested call for resampling. Exposed the API and tested with the existing call of get-neck-ik and get-ik-joint-positions in low-level.lisp

**Supported resampling axes** : (:x :y :z :theta)

Sample call:
```
(setf ?test-pose (tf:make-pose-stamped "triangle_base_link" 0.0 (tf:make-3d-vector 0.689 -0.2486 0.1648) (tf:make-quaternion 1 0 0 0)))

(setf ?seed-state (btr::make-robot-joint-state-msg (btr:get-robot-object) :joint-names '("left_arm_0_joint" "left_arm_1_joint" "left_arm_2_joint"
       "left_arm_3_joint" "left_arm_4_joint" "left_arm_5_joint"
       "left_arm_6_joint")))

(ik::find-ik-for (?test-pose "triangle_base_link" "left_arm_7_link" ?seed-state)
        (ik::with-resampling (0.1d0 :x 0.5d0 -0.5d0 0.1d0)
          (ik::with-resampling (0.0d0 :z 0.5d0 -0.5d0 0.1d0))))
```


cram_external_interfaces/cram_ik_interface/src/ik.lisp
```diff
@@ -134,3 +134,81 @@ If not valid solution was found, returns NIL."
...
+                                     (funcall pseudo-pose value)
+                                     base-link-evaled
+                                     tip-link-evaled
+                                     seed-state-msg-evaled))
+
+                              ;; When a solution is obtained parse accordingly
+                              (when (and solution-msg
+                                         (or (not solution-valid-p-evaled)
+                                             (funcall solution-valid-p-evaled solution-msg)))
+                                (return (values solution-msg
```
amar-fayaz, line 211: I just noticed that this line is not returning the value to the higher levels in the nest, resulting only the outermost nested value being returned. I am investigating into this.

amar-fayaz:
Added the support for resampling angles in the z axis and fixed the nesting output value issue

cram_external_interfaces/cram_ik_interface/src/ik.lisp
amar-fayaz: This is fixed in the new commit

gaya-:
It doesn't work on the PR2. There is a little hack:

https://github.com/cram2/cram/pull/171/files#diff-e615cfca9203aadbaa031d85654ee293R848

maybe it doesn't work with macros or so.

gaya-:
Solutions are (1) change the hack into something cleaner, probably something like how cram-tf:*robot-base-frame* is defined at startup-ros or (2) fix the problem on the macro side.

gaya-:
The issue doesn't seem to be in the name of the service. But PR2's IK service doesn't like the service call. Perhaps we should just switch onto our IK service for the PR2 as well...

gaya-:
@amar-fayaz I cannot figure out from the first glance what you've changed in the service call, so please take a look yourself and make sure that PR2's IK service is also working. If it is difficult to support PR2's IK service as well, we can consider switching PR2 to our kdl_ik_service.

amar-fayaz:
1. Fixed the offset issue
2. Added support for collision check with the resampled joints
3. Fixed the logging level clutter

gaya-:
@amar-fayaz I will wait for the second part of this functionality after all. You will most likely have to change the arguments of the collision check function and things might end up looking a bit different. So feel free to solve the next Trello card also in this branch under this PR.

amar-fayaz:
Updated to a working version of the base resampling.

1. A close inspection of the working of the resampling is still pending as per the chat we had
2. Added the option to disable logging. Some warning still keeps coming out
"Unknown tag "contact_coefficients" in /robot[@name='boxy_description']/link[@name='left_gripper_gripper_right_link']"
This is not from our python file. If it is accessible to us, please mention it so that I can fix it there as well.

amar-fayaz:
Updated with the latest review comments

gaya-:
So the bug comes from wrong coordinate frames :)

To understand the coordinate frames, one needs to understand one simple linear algebra concept:

if I have a pose in some frame, e.g., ``parent-T-child``, adding offsets to the translation of that pose is [equivalent to multiplying it with a translation transform **from the left**](https://en.wikipedia.org/wiki/Translation_(geometry)#Matrix_representation): ``parent-T-child_offsetted = parent'-T-parent * parent-T-child``. So our offsets are defined in the parent coordinate frame, i.e. the offsets will be aligned with the coordinate frame grid of the parent. Not the grid of the child frame, not the grid of the global fixed frame, but the parent frame. If we wanted to apply the offsets in the child frame, we would have to multiply **from the right**. But what we want here is an offset in the root frame, so multiply **from the left**.

The resampling that we are doing and the offsets that we are calculating, they are for the tip link of the kinematic chain, i.e. the tool frame, our ``offsetted-goal-pose``. The actual goal pose is a pose in torso frame: ``torso-T-tool``. The ``offsetted-goal-pose`` is as well: ``torso-T-tool_offsetted = torso'-T-torso * torso-T-tool``. So we are multiplying from the left and our ``:x, :y, :z`` offsets are in the torso coordinate frame. In fact, ``torso'-T-torso = [ I | (:x; :y; :z; 0)]``.

Similarly to offsetting the goal pose, we can apply the inverse of these offsets to the root link of our kinematic chain, i.e., to the torso link, to keep the IK solution constant. So, e.g., if we have ``map-T-torso``, we can apply the offset ``map-T-torso * torso-T-torso' = map-T-torso'``. The offsets are defined in the torso coordinate frame, so in this case we multiply **from the right**. And you notice that here we use ``torso-T-torso'`` and before we had the inverse: ``torso'-T-torso``.

To summarize, we either move the tool frame closer to the torso frame (by making little steps aligned with the torso coordinate frame grid), or we move the torso frame closer to the tool frame (again, in torso grid steps).

If your implementation, what happens is:
  * our inverse kinematics for ``torso-T-tool`` doesn't exist, so we have to resample
  * so, instead, we find the IK for ``torso'-T-tool``, for that
    * we come up with a ``torso'-T-torso`` by calculating the :x, :y, :z offsets as sampling values
    * we apply this offset to our goal: ``torso'-T-torso * torso-T-tool = torso'-T-tool``
    * we call the IK solver with the new ``torso'-T-tool``
  * then, because now we have a solution for ``torso'`` and not ``torso``, we correct the torso position:
     ``map-T-torso' = map-T-torso * torso-T-torso'``
    * to correct the Z: .

 ``map-T-base_offsetted = map'-T-map * map-T-base``. But we cannot just assume that ``map'-T-map`` is the same as ``torso'-T-torso``. The correct chain would be ``map-T-base * base-T-torso * torso-T-tool = map-T-base * base-T-torso * torso-T-torso' * torso'-T-tool`` In fact, this is only the case if the two coordinate frame orientations are perfectly aligned. If they're not, we observe a bug.

PR #174, opened 2020-05-25
[common-fail]Unit test for different location solution

amar-fayaz:
* Created unit tests for common-fail:next-different-location-solution used alone and along with common-fail:retry-with-loc-designator solutions. 
* Created the asd system.
* Parameterized the default distance filter threshold to reuse in testing.

Note: I wasn't able to find any issues of the robot getting stuck in the same place. Please check if I"m missing something here.

### DIFF
--- a/cram_common/cram_common_failures/cram-common-failures-tests.asd
+++ b/cram_common/cram_common_failures/cram-common-failures-tests.asd
@@ -31,36 +31,24 @@
   :license "BSD"
   
   :depends-on (lisp-unit
+               roslisp
+               cram-language
+               cram-prolog
+               cl-urdf
                cram-common-failures
                cram-bullet-reasoning
                cram-pr2-description
-               cram-robot-pose-gaussian-costmap
                cram-robot-interfaces
-               roslisp
-               cram-prolog
                cl-transforms
                cl-transforms-stamped
-               cl-urdf
-
-               cram-language
-               ;cram-executive
                cram-designators
-               ;cram-projection
-               ;cram-occasions-events
-               ;cram-utilities
-
                cram-object-knowledge
-               ;cl-bullet
-               ;cram-bullet-reasoning-belief-state
-               ;cram-bullet-reasoning-utilities
-
                cram-location-costmap
                cram-btr-visibility-costmap
                cram-btr-spatial-relations-costmap
                cram-occupancy-grid-costmap
-
-               
-               )
+               cram-robot-pose-gaussian-costmap)
+  
   :components
   ((:module "tests"
     :components
--- a/cram_common/cram_common_failures/cram-common-failures-tests.asd
+++ b/cram_common/cram_common_failures/cram-common-failures-tests.asd
@@ -1,19 +1,19 @@
-;;; Copyright (c) 2012, Amar Fayaz <amar@uni-bremen.de>
+;;; Copyright (c) 2020, Amar Fayaz <amar@uni-bremen.de>
 ;;; All rights reserved.
-;;; 
+;;;
 ;;; Redistribution and use in source and binary forms, with or without
 ;;; modification, are permitted provided that the following conditions are met:
-;;; 
+;;;
 ;;;     * Redistributions of source code must retain the above copyright
 ;;;       notice, this list of conditions and the following disclaimer.
 ;;;     * Redistributions in binary form must reproduce the above copyright
 ;;;       notice, this list of conditions and the following disclaimer in the
 ;;;       documentation and/or other materials provided with the distribution.
 ;;;     * Neither the name of the Intelligent Autonomous Systems Group/
-;;;       Technische Universitaet Muenchen nor the names of its contributors 
-;;;       may be used to endorse or promote products derived from this software 
+;;;       Technische Universitaet Muenchen nor the names of its contributors
+;;;       may be used to endorse or promote products derived from this software
 ;;;       without specific prior written permission.
-;;; 
+;;;
 ;;; THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 ;;; AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 ;;; IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
@@ -29,7 +29,7 @@
 (defsystem cram-common-failures-tests
   :author "Amar Fayaz"
   :license "BSD"
-  
+
   :depends-on (lisp-unit
                roslisp
                cram-language
@@ -48,7 +48,7 @@
                cram-btr-spatial-relations-costmap
                cram-occupancy-grid-costmap
                cram-robot-pose-gaussian-costmap)
-  
+
   :components
   ((:module "tests"
     :components
--- a/cram_common/cram_common_failures/cram-common-failures-tests.asd
+++ b/cram_common/cram_common_failures/cram-common-failures-tests.asd
@@ -1,0 +1,68 @@
+;;; Copyright (c) 2012, Amar Fayaz <amar@uni-bremen.de>
+;;; All rights reserved.
+;;; 
+;;; Redistribution and use in source and binary forms, with or without
+;;; modification, are permitted provided that the following conditions are met:
+;;; 
+;;;     * Redistributions of source code must retain the above copyright
+;;;       notice, this list of conditions and the following disclaimer.
+;;;     * Redistributions in binary form must reproduce the above copyright
+;;;       notice, this list of conditions and the following disclaimer in the
+;;;       documentation and/or other materials provided with the distribution.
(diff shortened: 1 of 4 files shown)


cram_common/cram_common_failures/cram-common-failures-tests.asd
```diff
@@ -0,0 +1,56 @@
+;;; Copyright (c) 2012, Amar Fayaz <amar@uni-bremen.de>
```
gaya-, line 1: wrong year

cram_common/cram_common_failures/cram-common-failures-tests.asd
```diff
@@ -0,0 +1,56 @@
...
+               cram-robot-interfaces
+               cl-transforms
+               cl-transforms-stamped
+               cram-designators
+               cram-object-knowledge
+               cram-location-costmap
+               cram-btr-visibility-costmap
+               cram-btr-spatial-relations-costmap
+               cram-occupancy-grid-costmap
+               cram-robot-pose-gaussian-costmap)
```
gaya-, line 50: Are all these dependencies really necessary? I doubt that. E.g., cram-language or cram-prolog, I don't see any prolog code or any CPL code.

cram_common/cram_common_failures/src/failure-handling-strategies.lisp
```diff
@@ -29,6 +29,10 @@
 
 (in-package :common-fail)
 
+(defparameter *default-distance-threshold* 0.05
+  "Distance threshold used as a default for identifying unique
```
gaya-, line 33: Why the line break? The line is not longer than 80 chars.

cram_common/cram_common_failures/src/failure-handling-strategies.lisp
```diff
@@ -91,7 +95,7 @@ after each iteration of the retry."
        (cpl:fail ,rethrow-failure))))
 
 (defun next-different-location-solution (designator
-                                         &optional (distance-threshold 0.05))
+                                         &optional (distance-threshold *default-distance-threshold*))
```
gaya-, line 98: The linebreak would probably be nice here, although I did not count the characters, so I wouldn't know.

cram_common/cram_common_failures/tests/different-location-solution-test.lisp
```diff
@@ -0,0 +1,100 @@
+;;;
+;;; Copyright (c) 2019, Amar Fayaz <amar@uni-bremen.de>
```
gaya-, line 2: wrong year

cram_common/cram_common_failures/tests/different-location-solution-test.lisp
```diff
@@ -0,0 +1,100 @@
...
+;;; CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+;;; SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+;;; INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+;;; CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+;;; ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+;;; POSSIBILITY OF SUCH DAMAGE.
+
+(in-package :cram-common-failures-tests)
+
+(defparameter *target-location* (cl-tf:make-pose-stamped
```
gaya-, line 32: Please don't use ``cl-tf``. It is shorter than ``cl-transforms`` and much shorter than ``cl-transforms-stamped`` but ``make-pose-stamped`` is from ``cl-transfroms-stamped`` package, not ``cl-tf``, and ``make-3d-vector`` is from ``cl-transforms``. Using ``cl-tf`` introduces a dependency on ``cl-tf``, which implements TF tree caching, and we don't need that here. Using ``cl-tf`` as a namespace will be removed in version 1.0 of CRAM, such that using it for symbols from other packages won't be possible anymore. And, yes, I understand that the autocompletion suggests ``cl-tf`` for these things but I don't want to introduce the dependency on ``cl-tf`` where it is not necessary.

cram_common/cram_common_failures/tests/different-location-solution-test.lisp
```diff
@@ -0,0 +1,100 @@
...
+                                 "map"
+                                 0.0
+                                 (cl-tf:make-3d-vector 0 1 0)
+                                 (cl-tf:make-quaternion 0 0 0 1)))
+
+(defparameter *test-location* (let ((?target *target-location*))
+                                (desig:a location
+                                         (reachable-for pr2)
+                                         (location (desig:a
+                                                    location
```
gaya-, line 42: weird linebreak, ``(desig:a location`` would read easier.

cram_common/cram_common_failures/tests/different-location-solution-test.lisp
```diff
@@ -0,0 +1,100 @@
...
+
+(defun setup-world ()
+  (setf rob-int:*robot-urdf*
+        (cl-urdf:parse-urdf (roslisp:get-param "robot_description")))
+  (prolog:prolog
+   `(and (btr:bullet-world ?world)
+         (rob-int:robot ?robot)
+         (assert (btr:object ?world :urdf ?robot ((0 0 0) (0 0 0 1))
+                             :urdf ,rob-int:*robot-urdf*))))
+  (clrhash (btr::get-updated-attachments))
```
gaya-, line 53: I don't think this is necessary. ``detach-all-objects`` should do the job. If it doesn't, please let me know, so I complain to @tlipps .

cram_common/cram_common_failures/tests/different-location-solution-test.lisp
```diff
@@ -0,0 +1,100 @@
...
+
+(define-test next-different-location-solution-test
+  (unless (eq roslisp::*node-status* :RUNNING)
+    (roslisp-utilities:startup-ros))
+  (setup-world)
+  (let ((first-location (desig:reference *test-location*))
+        (new-location (desig:reference
+                       (common-fail:next-different-location-solution *test-location*))))
+    (assert-true (< cram-common-failures::*default-distance-threshold*
+                    (cl-tf:v-dist
```
gaya-, line 67: The ``cl-tf`` thing.

cram_common/cram_common_failures/tests/different-location-solution-test.lisp
```diff
@@ -0,0 +1,100 @@
...
+                              common-fail:looking-high-level-failure
+                              common-fail:perception-low-level-failure) (e)
+                           (common-fail:retry-with-loc-designator-solutions
+                               *test-location*
+                               robot-location-retries
+                               (:error-object-or-string e
+                                :warning-namespace (fd-plans search-for-object)
+                                :rethrow-failure 'common-fail:object-nowhere-to-be-found) 
+                             (push (desig:reference *test-location*) location-output))))
+
```
gaya-, line 90: So many newlines o_O?

cram_common/cram_common_failures/tests/different-location-solution-test.lisp
```diff
@@ -0,0 +1,100 @@
...
+
+                      (cpl:fail 'common-fail:looking-high-level-failure))))
+    (assert-true  (eq 8 (length location-output)))
+    (maplist (lambda (pose-list)
+               (when (>= (length pose-list) 2)
+                 (assert-true (< cram-common-failures::*default-distance-threshold*
+                                 (cl-tf:v-dist
```
gaya-, line 97: ``cl-tf`` thing

gaya- (changes requested):

gaya-:
Thanks for the PR. I have a request for minor changes. They shouldn't take more than 5 minutes.

cram_common/cram_common_failures/cram-common-failures-tests.asd
amar-fayaz: both cpl and prolog are being used in [line 79](https://github.com/cram2/cram/pull/174/files#diff-22b1f4fa4877bdc39600856b196d41e3R79) and [line 48](https://github.com/cram2/cram/pull/174/files#diff-22b1f4fa4877bdc39600856b196d41e3R48) of the test code.

I kept the other costmaps so as it stays as it is when we use the robot, in case any of them adds a change. All the other packages are used as well.

amar-fayaz:
Updated for changes as per the review.

cram_common/cram_common_failures/cram-common-failures-tests.asd
gaya-: indeed, my bad

gaya-:
Right now, the current-robot-pose generator is turned off:

https://github.com/cram2/cram/blob/boxy-melodic/cram_common/cram_robot_pose_gaussian_costmap/src/current-pose-generator-and-pose-validator.lisp#L33

That's why one doesn't observe the unwanted behavior in the demo.

But even if the current pose generator is turned back on, your tests still pass, so it must be something else....

Thanks.

PR #175, opened 2020-05-28
Fix dual-arm attachments: loose stuff seems to be buggy

tlipps:
To check if the pose of an attached object was already updated, updated-link-in-attachment saves which links are attached and which were already updated by traversing the robots link tree. Since the object can be attached to both arms and there are motions which allow to move independently from each arm, it is not known if the pose of the attached object was already updated by the other arm. Because updated-link-in-attachment cannot know, if motions were executed for both links/arms the object is attached to, only links which are not loosely attached are updated. In particular, in the below code this resulted in checking if the traversed links for an attachment match with its list of links establishing a non loose attachment. 

Trello: https://trello.com/c/75sIYAGF/291-fix-dual-arm-attachments-loose-stuff-seems-to-be-buggy

### DIFF
--- a/cram_3d_world/cram_bullet_reasoning/src/robot-model.lisp
+++ b/cram_3d_world/cram_bullet_reasoning/src/robot-model.lisp
@@ -368,9 +368,11 @@ was updated by checking if `link' was already updated. The already
 updated links are saved under the attachment name in `updated-attachments'.
 If all links of an attachment were updated the entry under the attachment
 name in `updated-attachments' gets deleted."
-    (let ((links-attached-to (mapcar #'attachment-link (car (cdr attachment))))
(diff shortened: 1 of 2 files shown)


gaya-:
Merged manually, thanks.

gaya-:
@tlipps This broke the assembly demo :( Which makes sense, because it's impossible to know if when  robot places something on something else, if that should be a loose attachment like with a tray or if it should be a non-loose attachment like with the top wing on the upperbody. I guess we'll have to included that in the plan...

gaya-:
@tlipps oh, i forgot that we already have a unidirectional-attachment predicate in the knowledge base. Then it's even easier to fix, just some bug sitting somewhere...

PR #176, opened 2020-05-28
btr:robot-attached-objects-in-collision has a bug when floor is in collision

tlipps:
Fixed problems mentioned in [this](https://trello.com/c/re3uy1ez/267-btrrobot-attached-objects-in-collision-has-a-bug-when-floor-is-in-collision-removing-anything-that-has-attachments-is-incorrect) Trello Card. Added some tests which are executable with `(lisp-unit:run-tests :all :cram-bullet-reasoning-tests)` after loading the cram-pr2-pick-place package.


gaya-:
Merged manually, thanks.

PR #177, opened 2020-05-29
Collision Flag Bug With Attachments

tlipps:
There were false collision information saved as mentioned in [this](https://trello.com/c/EFbUJpwL/278-fix-pr-150-collision-flags-code) Trello card. This should be now fixed. Further tests covering this case were added:

- attach object a and b: when attaching object a and c, collision information is correct
- attach object a and b and attaching object a and c: when detaching object a and b, collision information is correct
- shopping demo scenario: attach basket to robot, attach object a to environment: simulate picking of a and placing of a in the basket by calling the attach- and detach-object methods in the sequence as specified by the events [object-attached-robot](https://github.com/cram2/cram/blob/boxy/cram_3d_world/cram_bullet_reasoning_belief_state/src/event-handlers.lisp#L87) and [object-detached-robot](https://github.com/cram2/cram/blob/boxy/cram_3d_world/cram_bullet_reasoning_belief_state/src/event-handlers.lisp#L381) in cram_btr_bf.

These test cases are written for static objects and for non-static objects.

### DIFF
--- a/cram_3d_world/cram_bullet_reasoning/cram-bullet-reasoning-tests.asd
+++ b/cram_3d_world/cram_bullet_reasoning/cram-bullet-reasoning-tests.asd
@@ -10,7 +10,7 @@
   :components ((:module "tests"
                 :components
                 ((:file "package")
-                 (:file "items-tests" :depends-on ("package"))
+                 (:file "items-tests" :depends-on ("package objects-tests"))
                  (:file "objects-tests" :depends-on ("package"))
                  (:file "robot-model-tests" :depends-on ("package"))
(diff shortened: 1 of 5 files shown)


cram_3d_world/cram_bullet_reasoning/cram-bullet-reasoning-tests.asd
```diff
@@ -10,7 +10,7 @@
   :components ((:module "tests"
                 :components
                 ((:file "package")
-                 (:file "items-tests" :depends-on ("package"))
+                 (:file "items-tests" :depends-on ("package objects-tests"))
```
gaya-, line 13: @tlipps You did not test this!!!!!

gaya-:
Merged manually, thanks.

PR #178, opened 2020-06-11
made spawn pose relative to wrist and set new arm configuration

Tigul:
# Changed
* Made the spawn position of the basket relative to l_wrist_roll_link 
* Set a new  joint configuration so fingers are perpendicular to the handle
* adjusted and refactored the demo to work with the new configuration 

## Screenshot
The screenshot shows the PR2 with its new configuration after finishing the demo.
PR2 with new configuration


cram_pr2/cram_pr2_shopping_demo/src/setup.lisp
```diff
@@ -188,22 +188,33 @@
...
-  (let ((basket-desig (desig:an object (type basket) (name b))))
-    (coe:on-event
-     (make-instance 'cpoe:object-attached-robot
-       :link "l_wrist_roll_link"
-       :not-loose t
-       :grasp :front
-       :arm :left
-       :object-name :b
-       :object-designator basket-desig))))
+   (urdf-proj:with-simulated-robot
```
gaya-, line 191: This will not work because the function is called at startup of a node, and you cannot perform stuff at this point yet...

gaya-:
@Tigul before making a PR, you should always restart your REPL and try with nothing loaded yet. A lot of times, what works in the current REPL state will not work, if you reload everything.

gaya-:
Merged manually, thanks!

PR #182, opened 2020-06-23
[Depends on PR #171] Collision free move-joints method 

amar-fayaz:
[PR 171: Resampling](https://github.com/cram2/cram/pull/171)
Wrote a wrapper function around the current move-joints method that invokes resampling of base and torso joint
Changes
1. The output of the current ```move-joints``` method is modified such that it outputs a true value upon completion. Some value other than nil is required since otherwise the resampling methods will consider the action as a failure.
2. Made a new method ```move-joints-avoiding-collision``` with the intended behavior of this PR

Notes:
1. I am not sure where we should be plugging the new method. So this PR does not include the deployment of the new method. But I have tested it.
2. Since a lot of methods are re-used and some values are required like pose for IK solving, I have used dummy values instead since they cannot be left nil (like left-tcp-pose and right-tcp-pose in ```perform-collision-check```, even though these values aren't used they are required to be non-nil for the collision check to happen). This does look hacky, but is unavoidable unless we're writing entirely new resampling and collision checks that fits for this use case.


amar-fayaz:
1. Updated the APIs to use the updated ones merged into boxy-melodic
2. Fixed the bug in the find-ik-for which didn't track all the joint value changes across different nested calls

amar-fayaz:
1. Created a new function to do resampling which doesn't use any functionality of the IK service.
2. Removed the clutter from the move-joints-avoiding-collision
3. Removed returning of the 't' from the original move-joints function as it isn't needed anymore.
4. Checks for either ik-solution-msg or joint-values instead of exclusively ik-solution-msg in ``perform-collision-check``
5. Removed the type check from the left-tcp-pose and right-tcp-pose from ``perform-collision-check`` as the values were only used as conditionals and the poses weren't used for anything else. This also enables the new move-joint function to use the collision check properly without having to pass dummy values.

PR #187, opened 2020-07-24
Giskard temporary merge

artnie:
**Giskard:**
- Outsource definition of cartesian goal from within giskard plans to *constraints.lisp*. 
- Instead of putting the cartesian goal into the *cartesian_constraints* field of the message, the cartesian goal is now defined as a json in the *constraints* field. This makes it easier to add more constraints to the goal, e.g. prohibit specific joints from moving, like the base or gripper-joint.
- Add giskard interfaces:
  - gripper by joint
  - neck by cartesian (look at point-of-interest)
  - neck by joint
- Remove "UpdateGodMap" entries in giskard messages. They didn't work for me (threw errors even) and I can work without them.
- Add *constraint* field to designators that move the arms. For now this field is only used to prohibit specific joints from moving.
- Outsource definition of collision-modes to *constraints.lisp* and split them into smaller modules. Having a collision constraint that avoids all collisions and another that allows specific collisions makes it is easier specify more elaborate definitions by combining them, as in :allow-hand and :allow-fingers, which generally avoid all collisions but allow it for specific links.

Example for constraining joints in their current position:
```
(let ((?pose some-pose)
      (?constraint '("left_gripper_joint" "triangle_base_joint" "odom_z_joint")))
  (perform
   (a motion
      (type moving-tcp)
      (left-pose ?pose)
      (constraints ?constraint))))
```

Move and keep joint to specific state during cartesian goal:
```
(let ((?pose some-pose)
      (?constraint '(("left_gripper_joint" 0.08) ("triangle_base_joint" -0.2))))
  (perform
   (a motion
      (type moving-tcp)
      (left-pose ?pose)
      (constraints ?constraint))))
```

**Related:**
- Adds joints and links for the neck and fingers to boxy description 
- Change *lifting* action to :allow-all collision, because giskard doesn't provide a straight movement while avoiding collision with nearby objects
- Wraps *moving-tcp* in unwind-protect
- Change cl-transforms and cl-tranforms-stamped to cl-tf2 in cram-giskard package, for a *lookup-transform* with static TFs
- Add cram-tf to cram-joint-states, since it is used there and probably forgotten


artnie:
@gaya- here is it. very, very conflicty

gaya-:
Merged manually. Thanks a lot!!
Pretty much merged everything, except the propagation of ``constraints`` property through the designator hierarchy. There is no use for this feature for now, so it's being left out.
You can track the current state of the giskard repo on the giskard branch.
When it's stable it will be merged back into devel aka boxy-melodic and boxy.

PR #188, opened 2020-07-24
[Bugfix][kdl-ik]Provided joint limits for continuous joints 

amar-fayaz:
There wasn't a category to represent continuous joints in the KDL enumerator. So whenever the joint limits were provided as [0.0, 0.0] it was considered as an immovable joint, thus failing IK.

Workaround as discussed, is providing 2pi limits on either side


gaya-:
Please only do the cropping for continuous joints. As of now, you simply change any joint that has 0.0 as limits to [-2pi; 2pi], if I got this right. Which means this also includes the fixed joints and all that. That would obviously be wrong.

Also, please change the limit from 2pi to 4pi as we discussed.

amar-fayaz:
> Please only do the cropping for continuous joints. As of now, you simply change any joint that has 0.0 as limits to [-2pi; 2pi], if I got this right. Which means this also includes the fixed joints and all that. That would obviously be wrong.
> 
> Also, please change the limit from 2pi to 4pi as we discussed.

[This line](https://github.com/cram2/cram/blob/eb647200dff49f1cb92ca466ced15ffefd0f8723/cram_external_interfaces/kdl_ik_service/src/kdl_ik_service/ik.py#L106) was already filtering out fixed joints from the list of joints where limits are calculated. But then I saw that I had missed taking the type of joint from the URDF rather than relying on KDL enumerator values. So that is fixed and a more cleaner solution is used now.

gaya-:
Seems fine but there is this super weird behavior that happens sometimes.

Try:

```lisp
(urdf-proj:with-simulated-robot
  (btr-utils:move-robot)
  (coe:on-event (make-instance 'cpoe:robot-state-changed))
  (exe:perform
   (desig:an action
             (type parking-arms)))
  (exe:perform
   (let ((?pose (cl-transforms-stamped:make-pose-stamped 
                  "map" 0.0
                  (cl-transforms:make-3d-vector 0.6 0.8 0.8)
                  (cl-transforms:make-identity-rotation))))
     (desig:a motion
              (type moving-tcp)
              (left-pose ?pose)
              (collision-mode :allow-all)))))
```

and right after

```lisp
(urdf-proj:with-simulated-robot
  (exe:perform
   (let ((?pose (cl-transforms-stamped:make-pose-stamped 
                  "map" 0.0
                  (cl-transforms:make-3d-vector 0.3 0.2 0.8)
                  (cl-transforms:make-identity-rotation))))
     (desig:a motion
              (type moving-tcp)
              (left-pose ?pose)
              (collision-mode :allow-all)
              (move-base t)))))
```

For me, it consistently goes into a weird pose. But when you call the last command the second time, it corrects itself.
Importantly, once it corrected itself, the solution will not fail anymore. It makes me feel like there is a caching mechanism of some sorts. Or there is some state in our Python code that makes it behave weirdly. So, you have to restart the launch file, restart the Lisp node (if you didn't have a ros core) and call the two commands above.

The weird behavior might not be connected to this PR but this is the first time I've seen it.

amar-fayaz:
The issue seems to be coming due to setting the limits to 4pi, -4pi. I have reduced it to 2pi, -2pi and things start working again. It is still weird why it didn't work with a bigger limit. Could this be due to the difference in how the joint values are applied in the FK in the python script and when the move-it code? (since you said there is a normalization happening, this could be the case. I did see values as high as 8.5 in the IK values for the continuous joints)

cram_external_interfaces/kdl_ik_service/src/kdl_ik_service/ik.py
```diff
@@ -103,10 +105,15 @@ def get_joint_limits_from_urdf(joint_name, urdf_obj):
         [joint_found] = [joint for joint in urdf_joints if joint.name == joint_name]
     except ValueError as e:
         raise ValueError("Error while trying to find joint named: %s in the urdf. Reason: %s" % (joint_name, e))
-    if joint_found.joint_type != 'fixed':
-        return joint_found.limit.upper, joint_found.limit.lower
-    else:
+    # Continuous joints (0 as joint limits will mess up the IK solution)
+    if joint_found.joint_type == 'continuous' and joint_found.limit.upper == joint_found.limit.lower == 0:
+        return 2 * math.pi, -2 * math.pi
```
gaya-, line 110: Actually, I dont understand why you have to specifically check if the limits are 0.
Shouldn't the limits of any continuous joint be -2pi; 2pi?
I tried this (just commented out the if of the limits == 0) and got the same weird behaviour of not converging, that you said was caused by -4pi; 4pi limits.
Then I changed the limits from -2pi; 2pi into -pi; pi, and the IK ended up having a lot of difficulty in finding solutions.

Why did I even play with this?
The IK for the right arm doesn't really work.
@Tigul  tried the mirror poses of the following:
https://hackmd.informatik.uni-bremen.de/lpAGpTLhQXWgIFIH4hDHow#

@amar-fayaz could you please take a look at this again.

PR #186, opened 2020-07-21
[spatial-cm] "Above" location keyword which supports dropping items

amar-fayaz:
[Trello](https://trello.com/c/oYrNo8Oo/213-dropping-action-desginator)

### Changes:
1. :above keyword for location designator. **:above** is similar to **:on** , except
     * :above has the option to mention a z-offset
     * using above with prismatic-containers will label the container as inaccessible. ie., using above for drawers in a transporting plan will open the drawer, but using it with a fridge will not open the fridge/oven.
2. Added conditions for spec:property, location-certain and location-accessible for **:above**

### Example:
```
(a location 
    (above (an object
                      (type drawer)
                      (urdf-name sink-area-trash-drawer-main)
                      (part-of iai-kitchen)))
   (z-offset 0.1)
   (side front)
   (side left)
   (range 0.2))
```
### Dropping items
Dropping items can be achieved by using the target location as :above locations
#### Example to test
In the package _cram_pr2_pick_place_demo_
```
(init-projection)
(spawn-objects-on-sink-counter)
(urdf-proj:with-simulated-robot
        (perform
         (an action
             (type transporting)
             (object (an object
                         (type bowl)
                         (location
                          (a location (on 
                                       (an object
                                           (type counter-top)
                                           (urdf-name sink-area-surface)
                                           (part-of iai-kitchen)))))))
             (location (a location (on (an object
                                           (type counter-top)
                                           (urdf-name sink-area-surface)
                                           (part-of iai-kitchen)))))
             (target (a location (above (an object
                                         (type drawer)
                                         (urdf-name sink-area-trash-drawer-main)
                                         (part-of iai-kitchen)))
                        (z-offset 0.1)
                        (side front)
                        (side left)
                        (range 0.2)
                        )))))
```

### Note
~~Noticed that in the current latest version of fetch-and-deliver-plans the drawer isn't sealed at the end of the transporting action. It seems like a bug with cpoe:location-reset. Apart from this, the sealing works fine and the object attaches propertly (tested in earlier version of boxy-melodic and also calling the sealing action separately after the transporting plan was done)~~ This was an issue in this PR and already fixed

PR #189, opened 2020-07-28
Fixing up Cleaning demo

amar-fayaz:
Depends on [Drop Action PR](https://github.com/cram2/cram/pull/186)

Using the `:above` keyword to place items in the trash and sink, simplified the existing cleaning-demo.
Removed the obsolete methods.


gaya-:
Merged manually. Thanks!

PR #190, opened 2020-07-31
[env-man] Incorporate reachability-orientation-offset predicate into costmaps.

cpollok:
Simple addition of the costmap:reachability-orientation-offset predicate into the environment mnaipulation costmaps.
If the fact is not defined for a robot, it just defaults to 0.0.
Tested with PR2 and Donbot.
Pr2 doesn't get offset, Donbot does.
Donbot doesn't work for other reason for now though.


gaya-:
Could you please try with the NIL instead of 0.0. So don't check in Prolog but in the Lisp function.

cpollok:
> 
> 
> Could you please try with the NIL instead of 0.0. So don't check in Prolog but in the Lisp function.

I think for that to work the predicate either needs to be defined for every robot or the default implementation can't default to (fail). Because when it fails the whole designator resolution should fail, right? While implementing I tested it without the check in Prolog and the costmap defaulted to the whole space in the kitchen (with the furniture cut out).
We could change the default implementation or make it necessary for every robot, but I didn't know if that was desirable.

PR #200, opened 2020-08-28
[fd-tests] Added infra and test cases for fd-plans

amar-fayaz:
Added test infrastructure for testing fetch and deliver plans.

All tests are set to pass now, since some tests that should not produce error causes error (eg: searching plan with second robot location) and blocks up test executions. 
Consider this PR as a review into the test infrastructure and polishing and finishing would be done in the Bugfix part of the Trello Card.


gaya-:
The delivering test crashed for me. But in general things look good, so I'm merging this one. And looking forward for the bugfix PR. Yay, PR/Issue #200. You should win some sweets or something for opening #200. :)

PR #201, opened 2020-09-01
Added move-base flag parsing for collision check

amar-fayaz:
[trello](https://trello.com/c/LDaau9Zq/366-in-check-collisions-of-picking-and-placing-actions-move-base-flag-is-not-being-considered)

Parsing move base flag for picking and placing actions to nil if the target is on the robot.

Note: I wasn't able to fully test this because I didn't manage to get donbot demo working on my system. I have only tested it discretely with my own action designators and explicit calls and would need some input in the actual scenario

### DIFF
--- a/cram_3d_world/cram_urdf_projection_reasoning/src/check-collisions.lisp
+++ b/cram_3d_world/cram_urdf_projection_reasoning/src/check-collisions.lisp
@@ -153,7 +153,23 @@ Store found pose into designator or throw error if good pose not found."
                         (right-lift-poses
                           (desig:desig-prop-value pick-up-action-referenced :right-lift-poses))
                         (object-name
-                          (desig:desig-prop-value object-designator :name)))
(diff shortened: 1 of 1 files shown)


cram_3d_world/cram_urdf_projection_reasoning/src/check-collisions.lisp
```diff
@@ -170,7 +186,7 @@ Store found pose into designator or throw error if good pose not found."
                           (cut:equalize-two-list-lengths left-poses right-poses)
                         (dotimes (i (length left-poses))
                           (urdf-proj::move-tcp (nth i left-poses) (nth i right-poses)
-                                               :allow-all)
+                                               :allow-al nil nil nil move-base)
```
gaya-, line 189: meh, you ate the second L. how did you even test this, even with your own designators...

PR #202, opened 2020-09-01
Assembly generic

artnie:
PR2, Donbot and Boxy were tested to do the assembly demo in projection. They all passed until the plane is lifted for assembling the wheels. No errors due to retracting from grasp occured. 
* Changes most of the assembly demo grasping trajectories. 
* Rotates the horizontal holder 90° c.-clockwise, therefore chassis, wheels and nuts needed repositioning
* Reduce visibility threshold for donbot, because the camera is occluded when he holds something
* Mirrors end effector link and tool frame to :right arm, because some plans had issues with not having both defined


artnie:
Donbots fingers are still too big for such intricate tasks as picking/placing bolts or the plane's window, but the projection allows it. The current implementation of grasp trajectories doesn't allow for robot-specific trajectories. Changing that would include some little but global changes to all trajectory definitions.

artnie:
@gaya- see if this fits your purpose, let me know if anything seems fishy.

gaya-:
Thanks. I removed the donbot description thingy, one-handedness was not the problem.

PR #205, opened 2020-09-08
added visualize-thinking function

Tigul:
# Added 
* A function which takes a robot-object and spawns the meshed of this object over the robot.

## Description 
The function creates a new object out of the meshes of the given robot-object and spawns it over the existing meshes. 
All mashes will be scaled by a factor of 1.3 additional the new meshes will be grey. The new object will be named :thinking. 

## Screenshot 
The thinker

### DIFF
--- a/cram_3d_world/cram_bullet_reasoning/src/items.lisp
+++ b/cram_3d_world/cram_bullet_reasoning/src/items.lisp
@@ -269,7 +269,7 @@ it is possible to change the pose of its attachments when its pose changes."
 
 (defmethod add-object ((world bt-world) (type (eql :mesh)) name pose
                        &key mass mesh (color '(0.5 0.5 0.5 1.0)) types (scale 1.0)
-                         disable-face-culling (compound *all-meshes-as-compound*) (collision-mask :default-filter) (collision-group :default-filter))
+                         disable-face-culling (compound *all-meshes-as-compound*) (collision-mask :all-filter) (collision-group :default-filter))
   (let ((collision-shape
           (etypecase mesh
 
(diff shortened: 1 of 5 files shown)


gaya-:
The screenshot looks good.
But the implementation will interfere with the projection reasoning, it will create unnecessary collisions.

  * first, please fix the indentation of your code and don't make lines longer than 80-90 characters
  * make sure the meshes move together with robot's body parts (attach the meshes to the robot)
  * then we need a function to remove the visualization, so please write that
  * then test if the robot gets into collisions by executing the pick and place demo (e.g. demo-random) and adding a call to your function and the cleanup function, for example here: https://github.com/cram2/cram/blob/boxy/cram_3d_world/cram_urdf_projection/src/low-level.lisp#L1168
at the beginning of this function and cleanup at the end and in the failure handling.
  * I assume that this will not work, so you'll need to set the collision flags of the meshes such that they are ignored in the collision check https://github.com/cram2/cram/blob/boxy/cram_3d_world/cram_bullet_reasoning_belief_state/src/belief-state.lisp#L95
  * then we need to add these functions everywhere, where thinking happens, but this we can do as a second step: https://github.com/cram2/cram/blob/boxy/cram_3d_world/cram_urdf_projection/src/low-level.lisp#L100  https://github.com/cram2/cram/blob/boxy/cram_3d_world/cram_urdf_projection/src/low-level.lisp#L778 https://github.com/cram2/cram/blob/boxy/cram_3d_world/cram_urdf_projection/src/low-level.lisp#L1001
  * keep in mind that visibility checks might not work with this mesh around the head, in which case move the meshes below the ground when rendering visibility masks: https://github.com/cram2/cram/blob/boxy/cram_3d_world/cram_bullet_reasoning/src/visibility-reasoning.lisp#L85
  * make the object color in the placing stability check the same as robot color (use a defparameter to define the color): https://github.com/cram2/cram/blob/boxy-melodic/cram_3d_world/cram_urdf_projection_reasoning/src/check-collisions.lisp#L359

  * test in assembly-demo (cram_projection_demos) with PR2
  * test assembly-demo with Boxy
  * test retail demo with Donbot

gaya-:
Decided to go with RViz visualizations instead. It is much more straightforward and easier for the user to understand.

PR #211, opened 2020-09-19
Universal costmaps and hack for grasping horizontal door handles

cpollok:
First of all: Sorry, you will have to merge this manually, because I'm not sure what needs to be done there.

This PR changes the way costmaps for doors are made. In general: It uses aabbs to cut out the parts too close to the door. The padding costmap parameter/prolog predicate is also considered.
There is some leftover code commented out for now. If you want you can of course delete it.

I tested this with donbot and PR2. But for some reason the costmap for the dish-washer-drawer-tray is broken for me (as in all of the kitchen is the costmap) and I couldn't get it to work.

Additionally there's a name-based hack to get the gripper rotated correctly for the dish-washer.


gaya-:
Cool, thanks. Looks like this is going to be sufficient to get things working for now. I'll take care of the merge, sure, no problem.

PR #213, opened 2020-09-24
Fix distance clipping in goals

cpollok:
Add distance clipping to the absolute distances given to container-state goal.
Change handling of :open distance, so it is just the opposite of :closed. As in "is the joint at a state more than <delta> greater than it's lower-limit?".

Also adds a small note at another point in the code to point out that it handles relative distances instead of absolute ones. If this is unwanted the commit can be easily omitted.


gaya-:
These 2 commits were manually cherry-picked:
https://github.com/cram2/cram/commit/73f3bc3a55e722cedc8d809285afbd582c9a11b7

I wonder why GitHub didn't understand this automatically...

PR #220, opened 2020-10-20
Consider indirect attachements between items in btr:robot-attached-objects-in-collision

tlipps:
[Trello](https://trello.com/c/tCCJBPI5/383-attachments-in-cramprojectiondemos-assembly-demo-when-we-pick-up-the-plane-a-bolt-is-in-collision-with-the-chassis-actually-they)


gaya-:
Cherry-picked into boxy-melodic. (Otherwise would have to revert the revert of the revert of ... with the boxy branch.)

gaya-:
Cherry-picked into boxy-melodic. (Otherwise would have to revert the revert of the revert of ... with the boxy branch.)

PR #227, opened 2021-08-02
Neem melodic

seba90:
I updated some stuff on my side, so I can finally generated NEEMs on melodic. I created a PR hope this is fine. If you are not intending to include the newest neem generation code just let me know.


hawkina:
Which KnowRob version is the one you used for this?

PR #242, opened 2021-12-15
Add function to reset bullet window

artnie:
Adds convenient function to reset the bullet window when it's frozen. Kills the thread, closes the window and opens a new one.
Call from your repl with `(btr-utils:reset-debug-window)`

### DIFF
--- a/cram_3d_world/cram_bullet_reasoning_utilities/src/package.lisp
+++ b/cram_3d_world/cram_bullet_reasoning_utilities/src/package.lisp
@@ -36,6 +36,7 @@
   (:export
    ;; vis-tools.lisp
    #:visualize-designator-costmaps #:visualize-gripper
+   #:reset-debug-window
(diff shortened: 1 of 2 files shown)


sunava:
@artnie opened this to noetic in [PR249](https://github.com/cram2/cram/pull/249), thank you

gaya-:
Cool idea. Btw, you could define another method for close-window and call this function when the user presses the X button on the window: https://github.com/artnie/cram/blob/d0f32b72b564976f59024a8e81ea8c94555a68a4/cram_3d_world/cl_bullet_vis/src/bullet-world-window.lisp#L28

Now that I think about it, we looked through all the commits in cl_opengl since 16.04 and didn't find anything that could cause this window freezing. So it's probably the behavior of X and Wayland that changed. That means that the code in this file probably has a bug, so worth looking through it.

PR #249, opened 2022-02-21
Add function to reset bullet window

sunava:
Adds convenient function to reset the bullet window when it's frozen. Kills the thread, closes the window and opens a new one.
Call from your repl with (btr-utils:reset-debug-window)
copied the PR from here: [PR242](https://github.com/cram2/cram/pull/242)


artnie:
Is this done?
Edit: can we merge this? It's really minor but very helpful, to me at least.

PR #267, opened 2022-06-28
Add http rosinstall for docker script

artnie:
we need that for docker


gaya-:
(It's https actually, only dinosaurs use http nowadays. But I'm jussayin, everyone has port forwarding anyway.)

PR #271, opened 2022-07-27
update neem interface

NeumannM89:
Working neem generation with Knowrob commit 59baa3f6c856bd977b54dd51d35e028520aea77d


hawkina:
works for me for that KnowRob commit.

PR #272, opened 2022-07-28
[projection] urdf-bullet-projection-enfironment tf topic

hawkina:
Within cram-projection:define-rpojection-enfironment urdf-bullet-projection-environment in the cram-tf:make-tf-broadcaster "tf_projection" was hardcoded as the topic name. If I were to change  the cram-tf:*tf-broadcasting-topic* variable in my startup function, it would still get overwritten and disregarded. 
Changed urdf-bullet-projection-environment to read in the topic name from the *tf-broadcasting-topic*instead. It is also set to default of "tf_projection" within the file so unless states otherwise, it would publish to tf_projection like it used to, so it shouldn't break anything for anybody. 
This is important for NEEM generation, since only tf data is currently logged. (In the future logging both would be cool)

### DIFF
--- a/cram_3d_world/cram_urdf_projection/src/projection-environment.lisp
+++ b/cram_3d_world/cram_urdf_projection/src/projection-environment.lisp
@@ -29,7 +29,7 @@
 (in-package :urdf-proj)
 
 (defvar *last-timeline* nil)
-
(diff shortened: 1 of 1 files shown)


gaya-:
Can we remove this branch now? ^^

hawkina:
oh sorry. will do.

PR #273, opened 2022-07-28
Ci tests

artnie:
Test this with `rosrun cram_tests test.sh`, then go to `roscd cram_tests/reports` and check it out.


cram_3d_world/cl_bullet_vis/tests/bullet-world-gl-context-test.lisp
```diff
@@ -26,7 +26,7 @@
 ;;; ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 ;;; POSSIBILITY OF SUCH DAMAGE.
 
-(in-package :bt-vis-test)
+(in-package :cl-bullet-vis-tests)
```
gaya-, line 29: Hope you tested this part well. To ensure that nobody else uses those names? Please do grep to make sure, and if anyone does use the name, you can always add a nickname.

artnie: This is only the test-package, the real one is another. I checked it and it's fine. Would be pretty bad if someone used any implementation from a test package. I didn't touch the package's `:nickname`. The package names are corrected to their system name (no system name is changed only packages), so `depends-on` in *.asd should still work without worries.

PR #276, opened 2022-08-24
add neem-interface-path as variable

NeumannM89:
Add the path to the neem-interface.pl as variable in cases the neem interface is not on the same pc as knowrob e.g. docker container.


artnie:
worked with the docker container so it should work in general

PR #278, opened 2022-10-17
[objects] Fix unidirectional-attach overwrite

artnie:
The files for retail and household each define the fact group `attachment-knowledge`. They overwrite the other, depending on the order of compilation. This PR renames the fact groups.


hawkina:
tested with popcorn-demo (branch devel) and donbot retail demo. Works fine :)

PR #281, opened 2022-11-15
Testsuit, Report, CI script

artnie:
Adds a testsuit for automated CI build and tests


artnie:
Check out the end of the build process, step 'Show test results' should contain the test results.

This check will always pass, regardless of failed and error'd tests. Until we fixed the tests we can just enjoy having a test-report.

PR #283, opened 2023-02-07
Devel forced CI - PR #281 with synchronized devel

artnie:


artnie:
This is the same PR as #281 but synchronized with the force-updated cram2/devel branch.

Check out the end of the build process, step 'Show test results' should contain the test results.

This check will always pass, regardless of failed and error'd tests. Until we fixed the tests we can just enjoy having a test-report.

artnie:
Main changes of this PR

Add cram-tests package: Give it the name of a cram package that has tests and it will run them for you. In the end it formats the test-results into a nice report.

Add bash script `cram_common/cram_tests/scripts/test.sh` that has a list of asdf packages. It runs the tests for all of them and concatenates their reports into one file in `cram_common/cram_tests/reports/`.

Equalize test package names so cram-tests can work easily with them, e.g. `cram-process-module-tests` is renamed into `cram-process-modules-test`, because `cram-process-modules` is the main package, and the corresponding tests are in `*-test`.

Fix depends-on and package.xml references, so cram builds and runs out of the box

Add a github-action workflow that (1) installs ROS, (2) pulls CRAM dep repositories, (3) builds ROS workspace, (4) runs the test-script and produces a report. This workflow is triggered whenever a pull request is opened to the devel branch.

Caveat: The github workflow will always be successful when all tests terminate. It does not check the test-results to determine a successful or failed result. Check out the test results in the workflow triggered by this PR, next to *All checks have passed* in *Show all checks*, then *Details*, unfold *Show test results*.

